PR #58, opened 2020-11-17
Added custom_bug_codes and custom_non_bug_codes

### LINKED ISSUE #8: allow user to specify status codes for which bug buckets should be reported
The user may have special status codes (e.g. similar to assertions) that they'd like bug buckets reported on regardless of the context.  

Examples:
- User is running the RESTler smoke test, and wants all requests to succeed.  They want to consume bugs in the same way as any other bug buckets.  In this case, they would specify 40* status codes in addition to the defaults.

rifiles:
Added custom_bug_codes and custom_non_bug_codes to restler settings and HttpResponse has_bug_code logic.

## PR Checklist
* [x] Applies to work item: Closes #8
* [x] CLA signed.
* [ ] Tests added/passed.  

## Info on Pull Request

* If custom_bug_codes list is specified in settings file, any status codes specified in the list will trigger bugs if received from the service in test.
* If custom_non_bug_codes list is specified, all status codes that are _not_ specified in this list will be trigger bugs.
* custom_bug_codes and custom_non_bug_codes list must be mutually exclusive - if both are specified then an error is raised.
* The * and ? wildcards can be used when specifying status codes
* Added new settings to SettingsFile.md

## Validation Steps Performed

cd restler-fuzzer/restler
python -m pytest ./unit_tests

Ran quick-start script

Tested on unit test server with:
* No lists specified
  * Passed with no change
* Both lists specified
  * Error handled
* Non-string type specified in list
  * Error handled
* custom_bug_codes with various codes specified
  * Successfully logged bugs included in list and 500 bugs
* custom_non_bug_codes with various codes specified
  * Successfully logged any code not in the list
* custom_non_bug_codes with '500' in the list
  * Ignored 500 code
* Tested wildcard handling for * and ?

### DIFF
--- a/docs/user-guide/SettingsFile.md
+++ b/docs/user-guide/SettingsFile.md
@@ -30,6 +30,14 @@ Checkers' Friendly Names:
 ### custom_bug_codes: list(str)
 List of status codes that will be flagged as bugs.
 
+Note:
+
+Use wildcard '\*' to allow any value after the star to exist.
+Ex: '2*' will match 200, 201, 210, etc.
+
+Use wildcard '?' to allow any value in that specific location.
+Ex: '2?1' will match 201 or 211, but not 202.
+
 ### custom_checkers: list(str) (default None)
 List of paths to custom checker files that will be loaded during runtime.
 
@@ -37,6 +45,14 @@ List of paths to custom checker files that will be loaded during runtime.
 A list of "non-bug" status codes.
 When this setting is defined,
 any status code received from the service-in-test that was _not_ included in this list will be flagged as a bug.
+
+Note:
+
+Use wildcard '\*' to allow any value on or after the star to exist.
+Ex: '2*' will match 200, 201, 210, etc.
+
+Use wildcard '?' to allow any value in that specific location.
+Ex: '2?1' will match 201 or 211, but not 202.
 
 ### dyn_objects_cache_size: int (default 10)
 Max number of objects of one type before deletion by the garbage collector
--- a/docs/user-guide/SettingsFile.md
+++ b/docs/user-guide/SettingsFile.md
@@ -27,8 +27,16 @@ Checkers' Friendly Names:
 * PayloadBody
 * Examples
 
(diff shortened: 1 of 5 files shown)


marina-p:
Please also add these to the settings json that contains all settings

docs/user-guide/SettingsFile.md
```diff
@@ -27,8 +27,16 @@ Checkers' Friendly Names:
 * PayloadBody
 * Examples
 
-### custom_checkers: str (default None)
-Path to a custom checker file that will be loaded during runtime.
+### custom_bug_codes: list(str)
+List of status codes that will be flagged as bugs.
+
```
marina-p, line 32: Please provide an example that also shows the regex format (right now that is not documented, as far as I can tell)

restler/engine/transport_layer/response.py
```diff
@@ -115,7 +117,19 @@ def has_bug_code(self):
 
         """
         if self._status_code:
-            return self._status_code.startswith('5')
+            if Settings().custom_non_bug_codes:
+                # If the non-bug-code list is nonempty,
+                # return False only if the status code exists in the list.
+                for code in Settings().custom_non_bug_codes:
```
marina-p, line 123: suggest rephrasing 'All codes except the ones in that list should be flagged as bugs.  Hence, return False ...'

marina-p (changes requested):
:clock1:

marina-p:
@rifiles I believe this one is not resolved yet

---
In reply to: [728684979](https://github.com/microsoft/restler-fuzzer/pull/58#issuecomment-728684979) [](ancestors = 728684979)

rifiles:
> @rifiles I believe this one is not resolved yet
> 
> In reply to: [728684979](https://github.com/microsoft/restler-fuzzer/pull/58#issuecomment-728684979) [](ancestors = 728684979)

You're right, missed that one.

marina-p (approved):
:shipit:

PR #64, opened 2020-11-17
Fix SDK requirement in README.

### LINKED ISSUE #60: dotnet error when running quick start
The quick start guide says I need an OpenAPI spec (which I have), and then shows a python command (not explicitly stated, but I have python). Running that command gives me this error:

```sh
$ python restler-quick-start.py --apiSpec ../openapi.yaml --restler_drop_dir .drop/
No executable found matching command "dotnet-/mnt/c/Users/aharper/workspace/api-v3/.restler/.drop/restler/Restler.dll"
```

This seems to imply there are additional steps to get going from scratch not described in the quick start guide. From looking at the readme, there is a "build" section which references dotnet, so I'm guessing that the build instructions are also a prerequisite not mentioned in the quick start. Is that right? Are there any other prerequisite steps to get started?

I recommend reviewing that doc from the perspective of a new user who isn't familiar with your project and has no additional context, and trying to follow your own directions and seeing if it works.

marina-p:
Closes #60

### DIFF
--- a/docs/user-guide/QuickStart.md
+++ b/docs/user-guide/QuickStart.md
@@ -1,7 +1,5 @@
 # Quick Start for Fuzzing with RESTler
 
-
-
 RESTler generates tests from your API specification,
 and executes these tests against your live service.
 In this quick start, you will first run RESTler with
@@ -24,7 +22,8 @@ The minimal requirements for fuzzing your API are:
 
 ## **Initial run**
 
-Once you have prepared (1) and (2), run the following command:
+Once you have built RESTler per the steps in the main README, and
+prepared (1) and (2), run the following command:
 
 ```python restler-quick-start.py --apiSpec <path to specification> --restler_drop_dir <path to RESTler drop>```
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/QuickStart.md
```diff
@@ -24,7 +22,8 @@ The minimal requirements for fuzzing your API are:
 
 ## **Initial run**
 
-Once you have prepared (1) and (2), run the following command:
+Once you have built RESTler per the steps in the main README, and
+prepared (1) and (2), run the following command:
 
 ```python restler-quick-start.py --apiSpec <path to specification> --restler_drop_dir <path to RESTler drop>```
```
rifiles, line 28: This should actually says api_spec_path instead of apiSpec - may want to add the fix to this PR.

rifiles: Nevermind - already fixed in main

marina-p: I need to rebase, that was already fixed by main

---
In reply to: [525434668](https://github.com/microsoft/restler-fuzzer/pull/64#discussion_r525434668) [](ancestors = 525434668)

rifiles (approved):

PR #66, opened 2020-11-17
Fixed bug in payload body checker that occurred if the request body did not start with a '{'.

rifiles:
Fixed bug in payload body checker that occurred if the request body did not start with a '{'.

The error occurred because substitute_body would return None if the '{' was not found and the payload body checker did not check for None prior to rendering it as a Request object.


marina-p (approved):
:shipit:

Issue #78, opened 2020-11-19
Create a pipeline to build a docker image


mgreisen:
The project https://github.com/microsoft/rest-api-fuz-testing needs a docker image to build on-top of. Create an image built on alpine linux. This should be build pulling from a tag or commit and saved into a test azure repository or pushed public when passing all testing stages.

marina-p:
@mgreisen publish-restler.yml will be run within the RAFT project, is that correct?

marina-p:
@mgreisen It's unclear from the description what change is needed to the existing RESTler CI pipeline.

mgreisen:
@marina-p 
The publish-restler.yml will be run as a pipeline in the ADO restler project. It will not be run as part of the RAFT project. 
The existing pipeline changes only include an extra uncompressed pipeline artifact drop.

marina-p:
Closing since the PR has been merged.

PR #82, opened 2020-11-20
New pipeline to publish restler docker image. #78

mgreisen:

We need to be able to create a public restler docker image. This PR adds a yaml file so that a pipeline that can be run from the private restler project as well as the dockerfile describing how the image is built. 

## PR Checklist
* [X] Applies to work item: #78 
* [X] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign.
* [ ] Tests added/passed.  
* [X] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different approach. Issue number where discussion took place: #xxx

## Info on Pull Request

*What does this include?*
Includes the yaml file, and the docker file.

The yaml file is setup to be able to create the restler image in a test repository that can be used for validation before publishing it publicly. 

There is also a change to the build yaml to create a drop that is used to pull data into the image. The existing drop contained a zip that I didn't want to touch. 

## Validation Steps Performed

*How does someone test & validate?*
Created a pipeline with the appropriate service connections that give the rights to the respective repositories. Created a test image and a production image (the production image is not published to docker hub yet) to verify the naming scheme. 
Run a RAFT BVT against the test image to verify it's constructed as we expected.

### DIFF
--- a/ci_build_pipelines/publish-restler.yml
+++ b/ci_build_pipelines/publish-restler.yml
@@ -1,0 +1,64 @@
+trigger:
+  batch : true
+  branches:
+    include:
+      - 'main'
+
+parameters:
+  - name: RestlerBuildTag
+    displayName: Restler Tag or Commit
+    default: "Tag Name"
+  - name: Production
+    displayName: Publish For Production?
+    default: No
+    values:
+    - Yes
+    - No
+
+variables:
+  - template: 'variables/version-variables.yml'
+  - name: version.revision
+    value: $[counter(variables['devRevision'], 1)]
+  - name: versionNumber
+    value: $(version.major).$(version.minor).$(version.revision)
+
+stages:
+  - template: stages/build/build.yml
+
+  - stage: publish
+    jobs:
+      - job: PublishDockerImage
+        pool:
+          vmImage: 'ubuntu-latest'
+        steps:
+        - checkout: self
+
+        - task: DownloadPipelineArtifact@2
+          inputs:
+            artifact: restler-fuzzer-drop-for-docker
+            path: $(Build.SourcesDirectory)/artifacts/restler-fuzzer-drop
+
+        - task: Docker@2
+          displayName: 'Publish Restler Test'
+          condition: and(succeeded(), eq('${{parameters.Production}}', 'No'))
+          inputs:
+              buildContext: $(Build.SourcesDirectory)/artifacts/restler-fuzzer-drop
+              command: buildAndPush
+              containerRegistry: acrrestler
+              repository: restler
+              Dockerfile: $(Build.SourcesDirectory)/docker/Dockerfile
+              tags: |
+                Test-$(Build.BuildNumber)
+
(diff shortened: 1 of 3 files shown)


stishkin (dismissed):

marina-p:
@ Thanks for adding this!  Since it's a non-trivial enhancement, could you please add a PR description with the pull request template here? [https://github.com/microsoft/restler-fuzzer/blob/main/docs/contributor-guide/PullRequestTemplate.md](https://github.com/microsoft/restler-fuzzer/blob/main/docs/contributor-guide/PullRequestTemplate.md)

marina-p (changes requested):
:clock1:

ci_build_pipelines/publish-restler.yml
```diff
@@ -0,0 +1,64 @@
+trigger:
+  batch : true
+  branches:
+    include:
+      - 'main'
+
```
marina-p, line 6: I just left a similar comment in the PBI - can you please help us understand when this is going to be triggered?

ci_build_pipelines/stages/build/steps/restlerfullbuild.yml
```diff
@@ -64,3 +64,7 @@ steps:
       artifactName: 'restler-fuzzer-drop-$(versionNumber)'
       publishLocation: 'Container'
 
+  - task: PublishPipelineArtifact@1
+    inputs:
+      targetPath: $(Agent.BuildDirectory)/restler_drop/
+      artifactName: restler-fuzzer-drop
```
marina-p, line 70: >restler-fuzzer-drop [](start = 20, length = 19)

Could you please name this 'restler-fuzzer-drop-for-docker'?  This makes it super clear that it's not the one to download if you just want a build of RESTler.

marina-p (dismissed):
:shipit:

ci_build_pipelines/publish-restler.yml
mgreisen: The pipeline would only be triggered manually. We only want to create new public images when you decide to release a new version.

bmc-msft: Instead of being manually triggered, you could use a different trigger.

This will run the workflow when a release has been published.
```
on:
  release:
    types: [published]
```

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release

ci_build_pipelines/stages/build/steps/restlerfullbuild.yml
mgreisen: Done.

ci_build_pipelines/publish-restler.yml
mgreisen: For now restler isn't using actions. It's something to consider though.

marina-p (approved):
:shipit:

PR #84, opened 2020-11-21
Added retry count for bugs that were not able to reproduce the first time

### LINKED ISSUE #72: Add retry count to bugs and always retry non-reproducible bugs
If a bucket is market as ‘non-reproducible’, any subsequent bug matching this bucket should be re-tried as well, and we should count these numbers of ‘reproducibility attempts/retries’ with each bucket.

Currently the bugs are tagged with whether or not they were reproduced, but there is no count for the number of times the bug was seen and if the bug was already found once, it currently will not attempt to reproduce again.

rifiles:
Added retry count for bugs that were not able to reproduce the first time

## PR Checklist
* [x] Applies to work item: Closes #72
* [x] CLA signed.
* [ ] Tests added/passed.  

## Info on Pull Request

* Added counts for number of reproduce retries
* Added counts for number of reproduce successes
* Will now replay bugs again if they were marked as not reproducible the first time
* Counts are printed to bug_buckets.txt
* Created new namedtuple and BugBucket class to improve readability of tuples

## Validation Steps Performed

cd restler-fuzzer/restler
python -m pytest ./unit_tests

Used debugger to force reproduce failures when running against unit test server (see below)

## Example:
In the below example, the bug failed to reproduce the first time, then it retried and succeeded twice.

InvalidDynamicObjectChecker_20x - Unable to reproduce bug - InvalidDynamicObjectChecker_20x_1.txt
Attempted to reproduce 3 time(s); Reproduced 2 time(s)
Hash: InvalidDynamicObjectChecker_20x_0c7bc79adedde76524a357aaa4f73dc367f6a19d
PUT /city/cityName HTTP/1.1\r\nHost: unittest\r\n\r\n{}\r\n
GET /city/_READER_DELIM_city_put_name_READER_DELIM HTTP/1.1\r\nHost: unittest\r\n\r\n

### DIFF
--- a/restler/engine/bug_bucketing.py
+++ b/restler/engine/bug_bucketing.py
@@ -14,6 +14,13 @@ class NewSingletonError(Exception):
 class UninitializedError(Exception):
     pass
 
+class BugBucket(object):
+    def __init__(self, sequence, reproducible, retry_count, reproduce_count):
+        self.sequence = sequence
+        self.reproducible = reproducible
+        self.retry_count = retry_count
+        self.reproduce_count = reproduce_count
+
 class BugBuckets(object):
     __instance = None
 
@@ -50,8 +57,8 @@ class BugBuckets(object):
 
         """
         # Iterate through each sequence in the bug buckets
-        for s in bug_buckets.values():
-            if s[0].last_request.hex_definition == sequence.last_request.hex_definition:
+        for b in bug_buckets.values():
+            if b.sequence.last_request.hex_definition == sequence.last_request.hex_definition:
                 return True
 
         return False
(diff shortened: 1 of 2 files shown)


restler/utils/logger.py
```diff
@@ -434,28 +436,30 @@ def add_hash(replay_filename):
...
+                if bug_bucket.reproducible:
+                    print(f'{name_header} - Bug was reproduced - {filename}', file=log_file)
                 else:
-                    name_header = f'{name_header} - Unable to reproduce bug - {filename}'
-                print(name_header, file=log_file)
-                print(f"Hash: {Bugs_Logged[bucket_hash][1]}", file=log_file)
-                sequence = seq_obj[0]
-                for request in sequence:
+                    print(f'{name_header} - Unable to reproduce bug - {filename}', file=log_file)
+                    print(f'Attempted to reproduce {Bugs_Logged[bucket_hash].reproduce_attempts} additional time(s); '
```
marina-p, line 459: >additional [](start = 97, length = 10)

Is 'additional' needed?

restler/engine/bug_bucketing.py
```diff
@@ -147,6 +154,41 @@ def _get_bug_hash(self, origin, sequence, hash_full_request, additional_str):
...
+        @param bug_code: The status code that was received when the bug was identified.
+        @type  bug_code: Str
+        @param bucket: Pointer to the bug bucket related to this bug
+        @type  bucket: OrderedDict - {seq_hex: BugBucket}
+        @return: Tuple containing whether the bug was reproduced,
+                 the new retry count, and the new reproduced count
+        @rtype : Tuple(bool, int, int)
+
+        """
+        if reproduce:
```
marina-p, line 173: >reproduce [](start = 11, length = 9)

This looks strange.  Suggest moving this out of the function, or making the reproduce parameter an integer, and handling >0 vs. 0 in the else.

marina-p (changes requested):
:clock1:

restler/utils/logger.py
rifiles: I added that because it looked a little weird and could be confusing, since the count only refers to attempts after the initial bug is logged.  For instance, if only the initial attempt fails and then we retry two more times it would say, "unable to reproduce bug", but then on the next line, "Attempted to reproduce 2 times; Reproduced 2 times."

If this looks ok I can get rid of it - or if all attempts (including the first attempt) should be part of the count, I can make that change, too.

PatGod: By definition, if the previous line says "unable to reproduce bug", then number_of_attempts > number_of_repros . Is this correct? 
If so, "additional" can be removed. OK?

rifiles: @patgod that's sort of not correct. If the first attempt was "unable to reproduce bug", but every attempt to reproduce the bug after that succeeds number_of_attempts == number_of_repros.  I can add the initial reproduce attempt that fails to the number_of_attempts count and then get rid of additional - probably makes the most sense

PatGod: @rifiles yes, the latter makes more sense (so that we always have number_of_attempts > number_of_repros after we print "unable to reproduce bugs"). Thanks!

marina-p (approved):
:shipit:

Issue #61, opened 2020-11-17
error when compiling (via quick start)


aharpervc:
After installing dotnet 3.1 sdk, rebuilding the program, and re-running the quick start script, I get this error:

```
$ python restler-quick-start.py --api_spec_path ../openapi.yaml --restler_drop_dir bin/
Log share was not specified.  Logs will not be uploaded.
Starting task Compile...

ERROR: Compiler failed. See logs in /home/aharper/restler_working_dir/Compile directory for more information. 

Task Compile failed.
Collecting logs...

ERROR: The RESTler grammar file path /home/aharper/restler_working_dir/Compile/grammar.py does not exist.
```

The log file:

```
Restler Information: 0 : Log share was not specified.  Logs will not be uploaded.
    DateTime=2020-11-17T15:53:06.3059771Z
Restler Information: 0 : Starting task Compile...
    DateTime=2020-11-17T15:53:06.3131157Z
Restler Information: 0 : Launching 'dotnet "/mnt/c/Users/aharper/workspace/api-v3/.restler/bin/restler/../compiler/Restler.CompilerExe.dll" "/home/aharper/restler_working_dir/Compile/config.json"'
    DateTime=2020-11-17T15:53:06.5334522Z
Restler Information: 0 : Process execution terminated in 00:00:00.7213000 with exit code 0x86: 'dotnet "/mnt/c/Users/aharper/workspace/api-v3/.restler/bin/restler/../compiler/Restler.CompilerExe.dll" "/home/aharper/restler_working_dir/Compile/config.json"'
    DateTime=2020-11-17T15:53:07.2638454Z
Restler Information: 0 : (7061) dotnet "/mnt/c/Users/aharper/workspace/api-v3/.restler/bin/restler/../compiler/Restler.CompilerExe.dll" "/home/aharper/restler_working_dir/Compile/config.json" exited with code: 134
    DateTime=2020-11-17T15:53:07.2706808Z
Restler Error: 0 : Compiler failed. See logs in /home/aharper/restler_working_dir/Compile directory for more information.
    DateTime=2020-11-17T15:53:07.2782879Z
Restler Information: 0 : Task Compile failed.
    DateTime=2020-11-17T15:53:07.2822032Z
Restler Information: 0 : Collecting logs...
    DateTime=2020-11-17T15:53:07.2879897Z
```

[...]

PR #86, opened 2020-11-22
Workaround for issue #61.

marina-p:
NSWag currently fails to parse a header definition with a boolean required property.

Work around this by removing the required.
There is currently no impact for RESTler correctness, because it does not
exclude optional properties.

### DIFF
--- a/src/compiler/Restler.Compiler/SwaggerSpecPreprocessor.fs
+++ b/src/compiler/Restler.Compiler/SwaggerSpecPreprocessor.fs
@@ -175,7 +175,16 @@ let rec private inlineFileRefs2
                                             if ref.Count() > 0 then
                                                 ref
                                             else
-                                                obj.Properties()
+                                                // TODO: Temporary workaround for issue #61 - NSWag failure to parse
+                                                // required boolean in Headers.  Remove this when the NSWag bug is fixed.
+                                                // The workaround is to remove the required property.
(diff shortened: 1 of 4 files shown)


src/compiler/Restler.Compiler/SwaggerSpecPreprocessor.fs
```diff
@@ -175,7 +175,17 @@ let rec private inlineFileRefs2
...
                                             else
-                                                obj.Properties()
+                                                // TODO: Temporary workaround for issue #61 - NSWag failure to parse
+                                                // required boolean in Headers.  Remove this when the NSWag bug is fixed.
+                                                // The workaround is to remove the required property.
+                                                // This is fine for now, since RESTler does not currently fuzz or
+                                                // learn from header values.
+                                                let headerInPath = sprintf "%s.%s" "headers" x.Name
+                                                if obj.Path.EndsWith(headerInPath) then
+                                                    //let headerName = obj. = obj.Path.Substring("headers." + "headers.".Length)
```
rifiles, line 185: Is this comment intentional?

marina-p: No, this needs to be removed. Thanks for catching

---
In reply to: [528958954](https://github.com/microsoft/restler-fuzzer/pull/86#discussion_r528958954) [](ancestors = 528958954)

rifiles (approved):

PR #90, opened 2020-11-25
Made it so compiler always generates refreshable token line and engine removes it if no token script specified.

### LINKED ISSUE #87: remove the compiler setting 'UseRefreshableToken'
The fact that this is a setting in the compiler causes the following two usability issues:

1) The compiled grammar from the quick start cannot be used, just adding authentication (user must find this in the setting, and set it in addition to setting the --token_refresh_command)

2) The grammar can't be easily used for parity testing with vs. without authentication.

## Proposal:

- Always generate the refreshable authentication token element in the grammar.
- If the ``--token_refresh_command`` is set, then invoke it and substitute the token
- If there is no token refresh command, the grammar element refreshable_authentication_token 'is ignored.

Always

rifiles:
Made it so compiler always generates refreshable token line and engine removes it if no token script specified.

## PR Checklist
* [x] Applies to work item: Closes #87
* [x] CLA signed.
* [x] Tests added/passed.  

## Info on Pull Request

* Compiler now ignore userefreshabletoken option and adds the token line in the grammar for all requests
  * UseRefreshableToken still accepted for backwards compatibility, but it is ignored
* Engine now deletes auth token lines if no auth token script specified in the settings file or as command-line argument
* Updated unit test grammar to include auth token lines

## Validation Steps Performed

cd restler-fuzzer/restler
python -m pytest ./unit_tests

Compiled and ran smoke test on NRP
Ran project unit tests in visual studio

### DIFF
--- a/ci_build_pipelines/stages/test/test.yml
+++ b/ci_build_pipelines/stages/test/test.yml
@@ -13,8 +13,14 @@ stages:
         steps:
           - template: steps/engineUnitTests.yml
 
-      - job: QuickStartTests
+      - job: QuickStartTests-Linux
         pool:
           vmImage: 'ubuntu-latest'
+        steps:
+          - template: steps/quickStartTests.yml
+
+      - job: QuickStartTests-Windows
+        pool:
+          vmImage: 'windows-2019'
         steps:
           - template: steps/quickStartTests.yml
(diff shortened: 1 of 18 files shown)


marina-p (approved):
:shipit:

ci_build_pipelines/stages/test/test.yml
```diff
@@ -13,8 +13,14 @@ stages:
         steps:
           - template: steps/engineUnitTests.yml
 
-      - job: QuickStartTests
+      - job: QuickStartTests-Linux
```
marina-p, line 16: >QuickStartTests [](start = 13, length = 15)

Thanks for adding this

marina-p (approved):
:shipit:

PR #91, opened 2020-11-30
Fixed bug that caused uuid4 suffix values to be misquoted

rifiles:
Fixed bug that caused uuid4 suffix values to be misquoted. Updated unit test grammar to test for this.

Before fix, when a uuid4 suffix type was quoted (i.e. in the body) it would quote the value from the dictionary and then append the uuid4 suffix as unquoted.

The bug also had the side effect of not matching uuid4 suffix values with their unquoted counterparts (i.e. from the path). Because the lookup was trying to match a quoted value with an unquoted value, they were treated as two different types.

This fix strips the quotes from the value before doing the comparison and adding of the suffix and then wraps the entire new suffix-added value with quotes when appropriate.


marina-p:
Could you please add to the description what the behavior was before vs. after the fix?  Otherwise, it's hard to tell from the description what bug you are fixing.

marina-p (approved):
:shipit:

PatGod (approved):

Issue #94, opened 2020-12-01
PayloadBodyChecker documentation


JoshuaEN:
Hello,

The PayloadBodyChecker appears to have a number of undocumented options (looking at the source code). It would be helpful to have these options documented, including the recipe file.

Use-case:

In my testing, it appears that the default settings for the payload body checker result in trying a low number of payload combinations, often ever sending a single value for most of the body parameters.

While it is often infeasible to try every possible combination, my thought it would be good to try a few fuzzed values for each parameter.

It appears to be possible to have the checker try more combinations via the settings. For example, [this recipe](https://github.com/JoshuaEN/restler-fuzzer-testing/blob/main/Server/restler-fuzzer/inputs/payload_body_recipe.json) does try other values for all of the body parameters, but it is too far in the other direction (making ~12000 requests vs. 58 with the defaults) and it could still just be exhaustively checking up to a (much higher) limit.

Thus, having the settings documented would be helpful.

Thanks!

PatGod:
@JoshuaEN Thank you for your feedback. The algorithms and techniques used in the PayloadBodyChecker are described in this paper: Intelligent REST API Data Fuzzing​​ (FSE'2020) https://patricegodefroid.github.io/public_psfiles/fse2020.pdf

Most of this checker's settings are discussed in that paper. 

Since there are many(!) possible setting configurations, the default settings roughly match what was found as the 'optimal' combination in the study described in that paper. But many other settings are indeed possible.

(Minimally, we should clarify the points above in the on-line documentation...)

Just out of curiosity, did you find more bugs with your custom recipe compared to the default one for your API/service under test? 

Thanks again!

JoshuaEN:
@PatGod Thank you for the link, it was very helpful in understanding not just what features are available, but why they are useful.

> Just out of curiosity, did you find more bugs with your custom recipe compared to the default one for your API/service under test?

No. I tried the default, the recipe I linked (which I now realize is rather flawed), and a few others built around the recommendations from that document with larger `value_max_combination` values. None found any bugs in the API I'm testing.

Having read the document, it is now my understanding that the payload body checker's intent is not to do valid value fuzzing (e.g. trying different valid values for an enum parameter), which is what I was trying to do.

Thanks again for your help!

Issue #95, opened 2020-12-01
Falsy PayloadBodyChecker settings are ignored when the default is Truthy


JoshuaEN:
Hello,

It appears that falsy PayloadBodyChecker settings are ignored when the default is truthy.

Example: Note how `size_dep_budget` is set to false in [engine_settings.json](https://github.com/JoshuaEN/restler-fuzzer-testing/blob/problem/payload-body-settings-ignored/Server/restler-fuzzer/results/Compile/engine_settings.json), it is loaded and reported as True in [PayloadBodyChecker.22992.txt](https://github.com/JoshuaEN/restler-fuzzer-testing/blob/problem/payload-body-settings-ignored/Server/restler-fuzzer/results/Fuzz/RestlerResults/experiment40856/logs/PayloadBodyChecker.22992.txt).

Reproduction steps for that specific run are here: https://github.com/JoshuaEN/restler-fuzzer-testing/tree/problem/payload-body-settings-ignored#this-run-can-be-reproduced-by

---

I think this is because the loaded setting is `or`ed with the default ([e.g.](https://github.com/microsoft/restler-fuzzer/blob/main/restler/checkers/payload_body_checker.py#L73)). Thus, for values that default to `true`, it is impossible to set them to `false`.

Note: I also verified that the settings were actually being loaded by changing the defaults for `_fuzz_valid` and `_fuzz_invalid` (in [payload_body_checker.py](https://github.com/microsoft/restler-fuzzer/blob/main/restler/checkers/payload_body_checker.py) ) to `False`, in the configuration setting `fuzz_valid` to `true`, and then confirming that the log reported `fuzzing valid True` and `fuzzing invalid False`.

Thanks!

PR #97, opened 2020-12-02
add note to payloadbodychecker doc

PatGod:
Add a note about the payloadbodychecker settings (related to issue #94)

### DIFF
--- a/docs/user-guide/Checkers.md
+++ b/docs/user-guide/Checkers.md
@@ -102,6 +102,21 @@ editing the format,
 changing object types,
 etc.
 
+__Note:__ The algorithms and techniques used in the PayloadBodyChecker
+are described in [Intelligent REST API Data Fuzzing​​](https://patricegodefroid.github.io/public_psfiles/fse2020.pdf) (FSE'2020).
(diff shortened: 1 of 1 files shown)


rifiles (approved):

PatGod:
Should this close issue #94? 

(As discussed, we will review separately the list of settings for this checker - should #94 remain open as a reminder for this @rifiles or can it be closed?)

rifiles:
> Should this PR close issue #94?
> 
> (As discussed, we will review separately the list of settings for this checker - should #94 remain open as a reminder for this @rifiles or can it be closed?)

We probably won't be updating the checker docs again until the recipe gets merged into the Settings file, correct? If so, the doc changes can probably be included in that PR, so I think it's safe to close the issue.

PatGod:
OK, will close #94 then. 

Note that the recipe file is not documented anywhere, and is obsolete with the settings file allowing checker settings, so the doc may not need to be updated any further. When we get to #95, a PR for #95 (if needed) will address this.

marina-p (approved):
:shipit:

PR #109, opened 2020-12-11
Updated driver output to show an engine failure if stderror is not empty

### LINKED ISSUE #107: Driver should output a failure when stderr is not empty
Currently the RESTler driver will output Task Succeeded at the end of a fuzzing/test run even if the run crashed and the stderr.txt file is not empty. This output should let the user know that there was an error, so they can investigate and not assume that the run completed successfully.

rifiles:
Driver will now check the stderror as well as the exit code from the engine when determining if the task succeeded. 

Closes #107

### DIFF
--- a/src/driver/Program.fs
+++ b/src/driver/Program.fs
@@ -251,9 +251,9 @@ module Fuzz =
                 None
         File.WriteAllText(workingDirectory ++ "EngineStdErr.txt", result.StandardError)
         File.WriteAllText(workingDirectory ++ "EngineStdOut.txt", result.StandardOutput)
-        if result.ExitCode <> 0 then
+        if result.ExitCode <> 0 || result.StandardError.Length <> 0 then
             Logging.logError <| sprintf "Restler engine failed. See logs in %s directory for more information. " workingDirectory
-        return result.ExitCode
+        return result
     }
 
     /// Runs the results analyzer.  Note: the results analyzer searches for all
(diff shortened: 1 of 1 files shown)


PatGod (dismissed):

src/driver/Program.fs
```diff
@@ -668,13 +668,19 @@ let main argv =
                         else
                             return None
                     }
+                    let exitCode =                         
+                        if result.ExitCode = 0 && result.StandardError.Length <> 0 then
+                            -1
```
marina-p, line 673: > -1 [](start = 27, length = 3)

Hi Richard, this seems confusing.  Does the engine swallow an exception and simply print the output to stderr?  If so, this should be an engine fix.  If not, what is the code path where this came up?   This change looks good to go in, this is just a clarification question.

marina-p, line 673: > -1 [](start = 27, length = 3)

Please define a constant that's not used in the engine (e.g. -2) and use it here, this will help flag such cases in telemetry.

rifiles: There was a crash in the fuzzing thread, which aborted that thread and printed to stderr.  When this happens the GC thread then cleans up and the main thread completes. This causes the engine to ultimately return a successful status code. 

The engine could (and probably should) keep track of thread exceptions and, if there were any present, return a non-successful error code. This quick fix seemed appropriate now because the driver probably still should report a failure if there was something in stderr for some reason.

I can create an issue for an engine fix to cure the root problem.

marina-p: Awesome, thank you, this sounds good.


---
In reply to: [541285192](https://github.com/microsoft/restler-fuzzer/pull/109#discussion_r541285192) [](ancestors = 541285192)

PatGod (approved):

PR #119, opened 2020-12-22
Cleaned up RestlerSettings set_arg calls.

rifiles:
Cleaned up RestlerSettings, so it is no longer required to make a set_arg call for every new settings arg.


marina-p (approved):
:shipit:

PR #120, opened 2020-12-25
Updated to allow handling of request and response bodies that start as arrays.

### LINKED ISSUE #67: Payload body checker can't handle bodies that don't start with '{'
Payload body checker can't handle bodies that don't start with '{'.

For instance, if a body is a list, which is still valid json, it will fail.

It fails first because it can't find the body (doesn't start with '{').
If the above is handled for '[' it will behave incorrectly because it assumes the body will always start with a key value pair, so the replacement bodies are created like this: "": [ "..." ] instead of simply [ "..." ]

rifiles:
Updated to allow handling of request and response bodies that start as arrays.

## PR Checklist
* [x] Applies to work item: Closes #67
* [x] CLA signed.
* [x] Tests added/passed.  

## Info on Pull Request

* Request and response parsers now recognize square brackets as valid body wrappers.
* Request schema parsing now treats arrays that are not key/value members correctly, which handles the cases where the body is an array instead of a dict. 
* Added new payload body checker test that handles an array body and a more complicated body (to be built upon and used more thoroughly later - specifically for improving bug bucketing)

## Validation Steps Performed

cd restler-fuzzer/restler
python -m pytest ./unit_tests

### DIFF
--- a/restler/engine/fuzzing_parameters/fuzzing_utils.py
+++ b/restler/engine/fuzzing_parameters/fuzzing_utils.py
@@ -157,13 +157,17 @@ def get_response_body(response):
     response = response.replace("\\'", "'")
     response = response.replace("\'", "'")
 
-    # extract the response body
+    # extract the response body by separating at the known delimiter
+    # Ex: METHOD /end/point HTTP/1.1 DELIM{}
     body_start = response.find(DELIM)
     if body_start == -1:
         return None
 
+    # Save the index where the body begins
+    body_start = body_start + len(DELIM)
+
     try:
-        start_char = response[body_start + len(DELIM)]
+        start_char = response[body_start]
     except Exception:
         return None
 
--- a/restler/engine/fuzzing_parameters/fuzzing_utils.py
+++ b/restler/engine/fuzzing_parameters/fuzzing_utils.py
@@ -151,16 +151,31 @@ def get_response_body(response):
     @rtype:  JSON
 
     """
+    from engine.transport_layer.messaging import DELIM
     # sanity cleanup
     response = response.replace('\\"', "'")
     response = response.replace("\\'", "'")
     response = response.replace("\'", "'")
 
     # extract the response body
-    body_start = response.find('{')
-    body_end = response.rfind('}')
+    body_start = response.find(DELIM)
+    if body_start == -1:
+        return None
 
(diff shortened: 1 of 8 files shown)


restler/engine/fuzzing_parameters/fuzzing_utils.py
```diff
@@ -151,16 +151,31 @@ def get_response_body(response):
...
     # extract the response body
-    body_start = response.find('{')
-    body_end = response.rfind('}')
+    body_start = response.find(DELIM)
+    if body_start == -1:
+        return None
+
+    try:
+        start_char = response[body_start + len(DELIM)]
+    except Exception:
```
marina-p, line 167: A short comment with an example would be useful (in particular, what DELIM vs. body_start are)

restler/test_servers/parsed_requests.py
```diff
@@ -19,11 +19,17 @@ def __init__(self, request_str: str, ignore_dynamic_objects=False):
...
         # Handle query string if necessary
         self.endpoint = endpoint_split[0].split('?', 1)[0]
-        header_body_split = endpoint_split[1].split('{', 1)
+        Auth_Token_Str = 'authentication_token_tag\r\n'
+        delim = DELIM
+        if DELIM in endpoint_split[1]:
+            pass
+        elif Auth_Token_Str in endpoint_split[1]:
+            delim = Auth_Token_Str
+        header_body_split = endpoint_split[1].split(delim, 1)
```
marina-p, line 28: An example would be useful here as well

restler/unit_tests/test_basic_functionality_end_to_end.py
```diff
@@ -291,6 +291,32 @@ def test_payload_body_checker(self):
         except TestFailedException:
             self.fail("Payload body failed: Garbage Collector")
 
+    def test_payload_body_checker_advanced(self):
+        """ This checks that the payload body checker sends all of the correct
+        requests in the correct order for more complicated bodies (including an array body)
```
marina-p, line 296: >in the correct order [](start = 17, length = 20)

Nice!  Could you please elaborate in this comment specifically what about orderings the new test covers?

marina-p (approved):
:shipit:

marina-p (approved):
:shipit:

PR #121, opened 2020-12-31
Improved payloadbody struct missing bug bucketing

rifiles:
The struct missing bucketization now adds {...} or [...] when an object or array branch is missing, so the user can more easily tell when a large branch is missing.

Also, a depth counter was added to the struct missing bucketization and a '+' for each layer of depth is prepended to any missing struct tag. 

**Examples:**
Default Schema:  {"level1": {"level21": {"level3": "val"}, "level22": "val"}}

* "level1" missing: StructMissing_/level1{...}
* "level21" missing: StructMissing_/+level21{...}
* "level3" missing: StructMissing_/++level3
* "level21" and "level22" missing: StructMissing_/+level21{..}/+level22

### DIFF
--- a/restler/engine/fuzzing_parameters/body_schema.py
+++ b/restler/engine/fuzzing_parameters/body_schema.py
@@ -20,6 +20,8 @@ class BodySchemaVisitor():
         # Can be used as an accumulator string while
         # traversing the body schema's params.
         self.val_str = ''
+        # Can be used to track the current node depth while traversing
+        self.depth = 0
 
 class BodySchema():
     """ Body Schema Class. """
--- a/restler/engine/fuzzing_parameters/request_params.py
+++ b/restler/engine/fuzzing_parameters/request_params.py
@@ -411,7 +411,9 @@ class ParamObject(ParamBase):
 
         """
         for member in self._members:
+            visitor.depth += 1
             member.check_struct_missing(check_value, visitor)
+            visitor.depth -= 1
 
     def _traverse(self, config: FuzzingConfig, func: str, accum_value):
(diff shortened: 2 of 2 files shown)


patricegodefroid:
Thanks Richard. Can you please confirm that this does NOT change the bucketization, but simply how buckets are identified?

Also, could you please include a concrete example here? (in addition to your abstract 'example' above...)

Thanks!

rifiles:
> Thanks Richard. Can you please confirm that this does NOT change the bucketization, but simply how buckets are identified?
> 
> Also, could you please include a concrete example here? (in addition to your abstract 'example' above...)
> 
> Thanks!

That's correct - no change to the bucketization - only the way it's printed.

Here is an example from the unit test server (using the new grammar from the other open PR):
Original Body:
```
{"properties":{"population":"0", "area": "5000","fuzzstring":true,"subproperties":{"subtest":true}}}
```

Bucket examples (not actually bugs in the unit test server, just spoofed them as bugs):
```       
StructMissing_/+population/+area/+strtest/+subproperties{...}
{'properties': {'fuzz': False}}

StructMissing_/properties{...}
{'fuzz': False}

StructMissing_/++subtest
{'properties': {'population': 0, 'area': 'fuzzstring', 'strtest': 'fuzzstring', 'subproperties': {'fuzz': False}}}

StructMissing_/+population/+area/+strtest/++subtest
{'properties': {'subproperties': {'fuzz': False}}}
```

PatGod:
OK, thank you for the clarifications! The new diagnosis/bucket-name looks good.

PatGod (approved):

PR #133, opened 2021-01-21
Certificate Based Authentication

landenblackmsft:
Enable Certificate Based Authentication.

Add support for user input --client_certificate_path
If available, add certfile to context via load_cert_chain.

### DIFF
--- a/build-restler.py
+++ b/build-restler.py
@@ -160,7 +160,7 @@ def publish_dotnet_apps(dirs, configuration):
         proj_file_path = dotnetcore_projects[target_dir_name]
         print(f"Publishing project {proj_file_path} to output dir {proj_output_dir}")
 
-        output = subprocess.run(f"dotnet restore {proj_file_path} --use-lock-file --locked-mode --force", shell=True, stderr=subprocess.PIPE)
+        output = subprocess.run(f"dotnet restore {proj_file_path} --use-lock-file --locked-mode --force -s https://api.nuget.org/v3/index.json", shell=True, stderr=subprocess.PIPE)
         if output.stderr:
             print("Build failed!")
             print(str(output.stderr))
--- a/docs/user-guide/SettingsFile.md
+++ b/docs/user-guide/SettingsFile.md
@@ -29,6 +29,7 @@ Checkers' Friendly Names:
 
 ### client_certificate_path: str (default None)
 Path to your X.509 certificate file in PEM format.
+
 If provided and valid, RESTler will attempt to use it during the SSL handshake.
 
 ### custom_bug_codes: list(str)
--- a/docs/user-guide/SettingsFile.md
+++ b/docs/user-guide/SettingsFile.md
@@ -27,6 +27,10 @@ Checkers' Friendly Names:
 * PayloadBody
 * Examples
 
+### client_certificate_path: str (default None)
+Path to your X.509 certificate file in PEM format.
+If provided and valid, RESTler will attempt to use it during the SSL handshake.
+
 ### custom_bug_codes: list(str)
 List of status codes that will be flagged as bugs.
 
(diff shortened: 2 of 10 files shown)


build-restler.py
```diff
@@ -160,7 +160,7 @@ def publish_dotnet_apps(dirs, configuration):
         proj_file_path = dotnetcore_projects[target_dir_name]
         print(f"Publishing project {proj_file_path} to output dir {proj_output_dir}")
 
-        output = subprocess.run(f"dotnet restore {proj_file_path} --use-lock-file --locked-mode --force", shell=True, stderr=subprocess.PIPE)
+        output = subprocess.run(f"dotnet restore {proj_file_path} --use-lock-file --locked-mode --force -s https://api.nuget.org/v3/index.json", shell=True, stderr=subprocess.PIPE)
         if output.stderr:
```
marina-p, line 164: Could you please either delete this before merging, or add a "--nuget_source" argument and pass it here?

src/driver/Program.fs
```diff
@@ -515,6 +516,11 @@ let rec parseEngineArgs task (args:EngineParameters) = function
         parseEngineArgs task { args with checkerOptions = args.checkerOptions @ [(checkerAction, specifiedCheckers |> String.concat " ")] } rest
     | "--no_results_analyzer"::rest ->
         parseEngineArgs task { args with runResultsAnalyzer = false } rest
+    | "--client_certificate_path"::certFilePath::rest ->
+        if not (File.Exists certFilePath) then
+            Logging.logError <| sprintf "The RESTler client certificate file path %s does not exist." certFilePath
```
marina-p, line 521: >RESTler [](http://example.com/codeflow?start=45&length=7)

remove 'RESTler'.

src/driver/Types.fs
```diff
@@ -65,6 +65,8 @@ type EngineParameters =
         /// Specifies whether to run results analyzer.
         runResultsAnalyzer : bool
 
+        /// File path to the client certificate.
+        certFilePath : string
```
marina-p, line 69: >string [](http://example.com/codeflow?start=23&length=6)

Could you please make this a string option, and set the default to None below?  Then, in Program.fs you can check for the None value so the parameter isn't sent to the engine at all when the user does not provide a cert.

src/driver/Program.fs
```diff
@@ -208,6 +208,7 @@ module Fuzz =
             sprintf "--restler_grammar \"%s\"" parameters.grammarFilePath
             sprintf "--custom_mutations \"%s\"" parameters.mutationsFilePath
             sprintf "--set_version %s" CurrentVersion
+            sprintf "--client_certificate_path \"%s\"" parameters.certFilePath
 
             (match parameters.refreshableTokenOptions with
```
marina-p, line 213: >refreshableTokenOptions [](http://example.com/codeflow?start=30&length=23)

See comment in Types.fs - could you please make this parameter only be passed if the user sets the option?  

match parameters.certFilePath with
| None -> ""
| Some cfPath -> <your current sprintf>

marina-p (changes requested):
:clock1:

docs/user-guide/SettingsFile.md
```diff
@@ -27,6 +27,11 @@ Checkers' Friendly Names:
 * PayloadBody
 * Examples
 
+### client_certificate_path: str (default None)
+Path to your X.509 certificate file in PEM format.
+
+If provided and valid, RESTler will attempt to use it during the SSL handshake.
```
landenblackmsft, line 33: Is this reference to "RESTler" okay, or should it be changed?

marina-p: Yes, this looks good.

marina-p (approved):
:shipit:

marina-p:
Thanks for contributing this feature!

marina-p (approved):
:shipit:

PR #137, opened 2021-02-02
Stream grammar to file. 

stishkin:
Make sure to discard preamble bytes since it trips up Python json module
- Streaming grammar is required to avoid OutOfMemory exception when processing large grammars

### DIFF
--- a/src/compiler/Restler.Compiler/Workflow.fs
+++ b/src/compiler/Restler.Compiler/Workflow.fs
@@ -65,7 +65,7 @@ let getSwaggerDataForDoc doc workingDirectory =
     {   swaggerDoc = swaggerDoc
         dictionary = dictionary
         globalAnnotations = globalAnnotations
-    }
+    }   
 
 let generateGrammarFromSwagger grammarOutputDirectoryPath swaggerDoc config =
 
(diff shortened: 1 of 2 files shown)


src/compiler/Restler.Compiler/Workflow.fs
```diff
@@ -67,6 +67,18 @@ let getSwaggerDataForDoc doc workingDirectory =
         globalAnnotations = globalAnnotations
     }
 
+type RESTlerFileStream(filePath, mode: System.IO.FileMode) =
```
marina-p, line 70: >RESTlerFileStream [](start = 5, length = 17)

If this is specifically for writing JSON that is read by the RESTler engine in Python, please put it under a module describing it as such and rename the class.  Right now it looks like this is generic and should always be used, given its location/name.

marina-p (changes requested):
:clock1:

src/compiler/Restler.Compiler/Workflow.fs
stishkin: This is a file stream implementation that removes preamble.

stishkin: Python JSON module fails to deserialize a file when preamble is added (or byte order mark as in wikipedia)

https://en.wikipedia.org/wiki/UTF-8

Byte order mark
If the UTF-16 Unicode byte order mark (BOM) character is at the start of a UTF-8 file, the first three bytes will be 0xEF, 0xBB, 0xBF.

The Unicode Standard neither requires nor recommends the use of the BOM for UTF-8, but warns that it may be encountered at the start of a file trans-coded from another encoding.[46] While ASCII text encoded using UTF-8 is backward compatible with ASCII, this is not true when Unicode Standard recommendations are ignored and a BOM is added. Nevertheless, there was and still is software that always inserts a BOM when writing UTF-8, and refuses to correctly interpret UTF-8 unless the first character is a BOM

marina-p: Ok.  Could you please move it to RESTler.Utilities, and add a comment that it should always be used, because files may need to be deserialized by the engine?   

---
In reply to: [569570570](https://github.com/microsoft/restler-fuzzer/pull/137#discussion_r569570570) [](ancestors = 569570570)

marina-p: also, suggest renaming to 'RestlerJsonFileStream'

---
In reply to: [569686943](https://github.com/microsoft/restler-fuzzer/pull/137#discussion_r569686943) [](ancestors = 569686943,569570570)

marina-p (approved):
:shipit:

Issue #138, opened 2021-02-03
Option to not use SSL certificate verification


miriver:
I propose that an option is added to use SSL without certificate verification.

I have my API service set up locally with a self-signed certificate. Unfortunately, Restler fails to make requests due to the certificate validation failing. Restler doesn't seem to have an option to use SSL but not validating the certificate, only to not use SSL at all.

I was able to fix this myself by changing line 50 in `messaging.py` from
`context = ssl.create_default_context()` to
`context = ssl._create_unverified_context()`.

If this is accepted, I can submit a PR to add an option to do this through the settings and/or command line arguments.

marina-p:
Hello Miguel,

This option should be supported.  It is already being discussed in #92.  

Would you mind testing if the update in branch rifiles/92 works for you, and if so send a PR or let me know it's good to merge?  It includes the change you proposed above, plus adding the option.

Thanks!

Marina

marina-p:
PR is merged, closing the issue.

Issue #92, opened 2020-12-01
Allow to disable SSL certificate validation


letmaik:
> 2020-12-01 11:19:31.668: 'Exception Creating Socket: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1108)'

I'm trying to test an SSL-only API locally where the certificate is self-signed. I couldn't find an option to disable certificate validation.

rifiles:
Hi Maik,

I believe the command-line option you're looking for is --no_ssl

letmaik:
```
no_ssl: bool (default False)

Set to True to disable SSL for requests
```

Doesn't sound like it to be honest. Are you sure?

rifiles:
I'm sorry, I believe you're correct - I answered too quickly. At the moment RESTler doesn't have an option to disable Python's default certificate validation. 

The entirety of the SSL code in RESTler can be found in messaging.py and the HttpSock constructor:

```
            self._sock = None
            host = Settings().host
            target_ip = self.connection_settings.target_ip or host
            target_port = self.connection_settings.target_port
            if Settings().use_test_socket:
                self._sock = TestSocket(Settings().test_server)
            elif self.connection_settings.use_ssl:
                context = ssl.create_default_context()
                with socket.create_connection((target_ip, target_port or 443)) as sock:
                    self._sock = context.wrap_socket(sock, server_hostname=host)
            else:
                self._sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
                self._sock.connect((target_ip, target_port or 80))
```

The ssl.create_default_context call performs the certificate checks by default.  I _think_ it should be possible to use ssl._create_unverified_context() or ssl._create_unverified_https_context() here, which hopefully could resolve your issue?

rifiles:
Hi Maik,

I made a quick update on branch rifiles/92.  Would you mind giving it a try on your API when you get a chance?

To disable you'll just need to add ```"disable_cert_verification": true``` to the settings file.

Thanks a lot

stishkin:
@letmaik - Are you running on Linux or Windows ? We are using RESTler on Linux container and using these instructions to add certificates to Linux store worked for us. 

https://askubuntu.com/questions/645818/how-to-install-certificates-for-command-line

letmaik:
[...]

PR #143, opened 2021-02-14
Adding disable cert validation option

miriver:
As discussed with @marina-p in #138, I verified this setting works for our scenario.
This was also discussed with @rifiles in #92.


marina-p (approved):
:shipit:

PR #170, opened 2021-03-29
Extend spec coverage file with concrete requests and responses.

marina-p:
This change adds an example of a concrete request for which the
data in the spec coverage file is reported.

The goal of this change is that information in the spec coverage file
can be used to troubleshoot failing requests using only:

- the spec coverage file
- backend logs for the service

The diff speccov script has been modified to avoid diffing the
concrete data that is now in the coverage file.

### DIFF
--- a/docs/user-guide/Testing.md
+++ b/docs/user-guide/Testing.md
@@ -71,6 +71,24 @@ Each request is represented by a hash of its definition.
         "status_code": "400",
         "status_text": "BAD REQUEST",
         "error_message": "{\n    \"errors\": {\n        \"id\": \"'5882' is not of type 'integer'\"\n    },\n    \"message\": \"Input payload validation failed\"\n}\n",
+        "sample_request": {
+            "request_sent_timestamp": null,
+            "response_received_timestamp": "2021-03-31 18:20:14",
+            "request_uri": "/api/blog/posts/5882",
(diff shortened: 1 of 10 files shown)


docs/user-guide/Testing.md
```diff
@@ -71,6 +71,7 @@ Each request is represented by a hash of its definition.
         "status_code": "400",
         "status_text": "BAD REQUEST",
         "error_message": "{\n    \"errors\": {\n        \"id\": \"'5882' is not of type 'integer'\"\n    },\n    \"message\": \"Input payload validation failed\"\n}\n",
+        "todo marina: update doc": ""
```
PatGod, line 74: Don't you want to finish this doc update as part of this PR?

PatGod: I mean: please eliminate this todo, or update the doc further if need be.

PatGod (approved):

docs/user-guide/Testing.md
marina-p: Oops, looks like this did not get committed.  Fixing.

---
In reply to: [604565291](https://github.com/microsoft/restler-fuzzer/pull/170#discussion_r604565291) [](ancestors = 604565291)

PatGod (approved):

Issue #175, opened 2021-04-01
add query parameter support  for "explode:false, style:form"


marina-p:
[Extracted from issue #153] 

I find that our server accepts arrays without square brackets and strings without quotes. The reason is that we are having style values of the following in our open API file:

explode: false
style: form
these two allow for csv format syntax for arrays without square brackets and without quotations for strings. See: https://swagger.io/docs/specification/serialization/

Is there an option in the compiler section of RESTler that we can turn-off square brackets in arrays and quotations around strings in the body parameters?
Could the examples file help us to mitigate the above body construction matter? At the moment, I have to manually go into grammar.py and edit the file. And it works only for GET requests only. PUT requests, I have to inspect again to see what is wrong. Is there a possibility that we can mitigate this matter with examples? If so, is it possible to provide a sample?
For example, this is a query that I have succeeded w/o any quotation around strings and brackets around array:

https://sample.com/test/site?groupCluster=KSLTJWY5&from=2021-01-26&to=2021-01-26&sort=api&product=microsoft&category=azure&severity=low&ids=1111111-1111-1111-1111-1111111111&fields=category,product,severity&pageSize=1000&pageFromKey=1&pageTotal=True

If I put brackets around the fields body parameter, which is an array, or if I put quotes around any of the strings, I get 400 errors.

In the OpenAPI, as an example, for fields parameter I have:

        - name: fields
          description: The fields to return the requested.
          in: query
          style: form
          explode: false
          schema:
            type: array
            items:
              type: string
I have 34/56 queries to have

          style: form
          explode: false
constructs in my OpenAPI specification file.

marina-p:
Closed by #181

PR #181, opened 2021-04-06
Fix query examples and support different serialization formats.

marina-p:

The primary purpose of this PR is to fix query parameter formatting and support two different serialization formats for queries: ```Form``` with ```explode: true``` or ```explode: False```.  This implements the format per the following spec: https://swagger.io/docs/specification/serialization/.

While implementing this, several issues with schema and example support for arrays were also fixed:

1) Multiple array values in examples are now supported (up to max 5).
2) Empty array examples in queries and bodies are supported.  These are represented with an array without a leaf element in grammar.json. 
 Previously, an empty array was represented as an object - this is now fixed, so it can be fuzzed correctly.

Closes issue #175. 

## PR Checklist
* [x ] Applies to work item: #175 
* [ x] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign.
* [ x] Tests added/passed.  
* [ x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different approach. Issue number where discussion took place: #xxx

## Info on Pull Request

- Changes to SwaggerVisitors.json to support the new style values and array fixes
- Updated and simplified logic in the CodeGenerator.fs to support different parameter serialization options
- New tests
 
## Validation Steps Performed

*Added new compiler test
*Manually tested that the engine works correctly with the payload body checker.  This will not be checked in because there is no way to automatically update the large grammar baseline in the engine tests.  This will be investigated separately.

### DIFF
--- a/src/compiler/Restler.Compiler/Compiler.fs
+++ b/src/compiler/Restler.Compiler/Compiler.fs
@@ -20,6 +20,8 @@ open Restler.Dictionary
 open Restler.Compiler.SwaggerVisitors
 open Restler.Utilities.Logging
 
+exception UnsupportedParameterSerialization of string
+
 module Types =
     /// A configuration associated with a single Swagger document
     type ApiSpecFuzzingConfig =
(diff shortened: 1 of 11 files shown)


src/compiler/Restler.Compiler/Compiler.fs
```diff
@@ -502,8 +516,8 @@ let generateRequestGrammar (swaggerDocs:Types.ApiSpecFuzzingConfig list)
         let orderedSwaggerDocs =
             swaggerDocs |> List.mapi (fun i sd -> (i, sd))
         let processed =
-            orderedSwaggerDocs.AsParallel()
-                              .AsOrdered()
+            orderedSwaggerDocs//.AsParallel()
```
PatGod, line 519: Does this need to be cleaned-up?

PatGod (approved):

PR #183, opened 2021-04-07
Support header parameters

### LINKED ISSUE #156: Support fuzzing request header
In some cases, we send some parameter in the request header, at the moment the compiler does not parse that, an example parameter:
   "parameters": [
          {
            "in": "header",
            "name": "serverComputerName",
            "description": "The name of the server computer (targetMachine).",
            "type": "string",
            "fullTypeName": "System.String"
          }
        ],

Can you please add support for this?

marina-p:
Changes include:
1. Fuzz header parameters of simple and array types.
2. Support restler_custom_payload_header in dictionary.  This works the
same way as the other types of custom payloads: if a parameter with the
same name is found in the spec, the custom payload from the dictionary
is used instead.
3. Fix bug in the earlier introduced feature that extra custom payload
headers (custom headers) can be added in the dictionary.  This allows the use
case of providing header parameters that are not included in the spec.

Testing:
- added compiler unit tests
- manual testing

Closes #156

### DIFF
--- a/docs/user-guide/FuzzingDictionary.md
+++ b/docs/user-guide/FuzzingDictionary.md
@@ -1,6 +1,6 @@
 # Fuzzing Dictionary
 
-The fuzzing dictionary allows users to configure sets of values for specific data types or individual parameters.  
+The fuzzing dictionary allows users to configure sets of values for specific data types or individual parameters.
 
 There are three categories of configurable values in the dictionary:
 
@@ -10,7 +10,7 @@ There are three categories of configurable values in the dictionary:
 
 
 
-RESTler *automatically* generates references to the dictionary when it compiles a Swagger specification.  The dictionary elements correspond to grammar elements in the RESTler grammar, and will determine part of a payload for one or more requests.  
+RESTler *automatically* generates references to the dictionary when it compiles a Swagger specification.  The dictionary elements correspond to grammar elements in the RESTler grammar, and will determine part of a payload for one or more requests.
 
 The following describes how each property in the dictionary is used in a RESTler grammar.  If needed, users may also customize the grammar further by referring directly to the dictionary.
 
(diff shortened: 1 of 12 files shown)


marina-p:
Still need to add documentation

PatGod (approved):

mbiuki:
Do I expect to see the actual value of the `uuid  `header type in `grammar.py`? Or should I expect to see the default value as indicated in `src/compiler/Restler.Compiler/Grammar.fs`: 
`PrimitiveType.Uuid, "566048da-ed19-4cd3-8e0a-b7e0e1ec4d72"`
They will be replaced w/ actual values of the header type that I have specified in `dict.json` within `restler_custom_payload_header ` attribute section, during run time?

marina-p:
The ```restler_custom_payload_header``` section needs to be specified at compile time.  Like all the custom payloads, when you specify it only at runtime, there is no matching grammar element and it will be ignored.  (Please let us know if you have feedback on this, ie. if you have a use case where you're not able to specify it at compile time)

So, if the grammar has ```restler_custom_payload_header```, there is no default value and the custom payload value will be used.  If the grammar does not have ```restler_custom_payload_header```, you should see a ```restler_fuzzable_uuid``` with the default value as in your message above (from the code).

mbiuki:
> The restler_custom_payload_header section needs to be specified at compile time. Like all the custom payloads, when you specify it only at runtime, there is no matching grammar element and it will be ignored. (Please let us know if you have feedback on this, ie. if you have a use case where you're not able to specify it at compile time)

A possible use case is if an attribute only needs to be added/dropped to/from `dict.json` during runtime w/o need to recompile. 

Compilation as far as I see it, comes in two phases:
1. Get the Compile folder and all the files in it by running `Restler.exe compile --api_spec openapi.yaml`. At this stage, if I have header attributes from the YAML file, they should be brought into `dict.json`. This is not the case at the moment.
2. After modifying `dict.json`, `restler_user_settings.json`, and `config.json`, go to Compile folder and recompile with `Restler.exe compile config.json`. After done so, I still do not see the correct values of header attributes in `grammar.py`.

marina-p:
The expected workflow is slightly different in RESTler at the moment than your step 1.  RESTler does not add header values from the specification into dict.json at compilation time, because it treats header parameters the same as any other parameter - generating a  ```restler_fuzzable_<type>(...)``` for their values in the grammar.  The only new keys based on the specification which RESTler will add to the dictionary is ```restler_custom_payload_uuid_suffix```.   So for headers, you need to explicitly add ```restler_custom_payload_header``` in the dictionary, and recompile again.

In your step 2, can you please confirm, did you add a new key-value pair in the ```restler_custom_payload_header``` section matching one of your header parameters, like below, recompiled, and in the grammar you still see a ```restler_fuzzable_guid```?  In the generated Compile directory, do you see this new ```restler_custom_payload_header``` section in the dictionary still?

```
  "restler_custom_payload_header": {
    "rating": [ "555", "999" ]
  }
```

mbiuki:
Hi @marina-p,

There are certain considerations from the design point of view that could help increase RESTler's autonomy. If decided to be implemented, it takes RESTler to the next level by helping users get up to speed with fuzzing more rapidly. Here are my suggestions:

1. For the first time that we run compile on `openapi.yaml` to populate JSON files such as `grammar` and `dict`, I suggest that the compiler populates the `dict.json` based on the parsing open API spec. This reduces complexity and chances for users to miss fixed body and header UUIDs. 
2. For the second time that the user runs compile command with giving the `config` file (assuming `config` file is populated properly), `dict.json` has already been set up properly, and the user only needs to populate the attributes.

PR #195, opened 2021-04-28
Added missing settings to doc: target_ip and target_port

c7h:
while testing, I realized that the documentation is missing entries for _target_port_ and _target_ip_. I've added these  to the docs.


marina-p (approved):
:shipit:

marina-p:
Thanks for updating!

PR #202, opened 2021-05-05
Working CBA changes

marina-p:
Creating PR for comparison with original cert_based_authentication changes.

working changes

Changes for CBA

remove cert password arg

refactor to new user args method

fixing merge issue


marina-p:
This PR is no longer needed

PR #206, opened 2021-05-11
New test mode for exhaustive regression testing.

marina-p:
This change implements the following customer request: 
in 'Test' mode, instead of stopping at the first
valid request, allow testing all available parameter combinations.

Specific changes:
- Integrate directed smoke test implementation into the main
fuzzing loop, because the majority of code from the fuzzing loop 
needed to be reused

- Introduce new option 'test_all_combinations', which executes every
rendering of the sequences used in the directed smoke test.  The
existing directed smoke test works as before, stopping at the
first valid rendering.

- Incrementally log spec coverage (also a customer request).

Testing:
- demo_server in both modes
- create_once testing
- unit test updates
- manual testing for various failure scenarios to make sure spec coverage is logged


PatGod (dismissed):

PatGod (approved):

marina-p:
@PatGod I have addressed the remaining TODOs and updated the engine unit tests.  Could you please review again? (There are significant changes from the previous iteration). 

For the tests: I decided that for this initial iteration I don't want to update baselines, but change the logic of the tests to make it obvious what the change in behavior is (and pass validation this way).  Specifically, I modified the tests to check that the baselines only differ in the *ordering* of sequences, but the sequences are still identical.  I left a TODO in the code to update baselines once this change is integrated.  (I also want to refactor these tests at some point to be easier to update, but this should not block this change).

restler/engine/core/driver.py
```diff
@@ -344,222 +413,10 @@ def add_producers(req, req_collection, req_list, dfs_stack):
 def compute_request_goal_seq(request, req_collection):
     req_list = add_producers(request, req_collection, [], [])
     if req_list == [] :
-        logger.write_to_main(f"\nSkipping request {request.method} {request.endpoint}\n", True)
+        logger.write_to_main(f"\nSkipping request {request.method} {request.endpoint}"\
+                             f"because a sequence satisfying all dependencies was not found.\n", True)
     return req_list
 
-
-def generate_sequences_directed_smoketest(fuzzing_requests, checkers):
```
PatGod, line 351: There might be a major design flaw introduced with these changes here.

The goal of the *directed* smoke test is to provide a way to test every request *once* with a cost *linear* in the number of requests in an API. This code guarantees that, for each request, a *single* sequence leading to the target request is selected, and then the target request is executed.

By eliminating this code, we go back to a *non-linear* algorithm like the previous/old smoke test (where multiple sequences corresponding to different interleavings of the partial order of dependency relations may be executed for a same target request).

In other words, this change undoes the very purpose of introducing a *directed* smoke test in the first place: with this change, the smoke test is no longer *directed* (like before the directed smoke test was implemented).

Please revert that part of this PR.

PatGod (changes requested):

restler/engine/core/driver.py
```diff
@@ -344,222 +413,10 @@ def add_producers(req, req_collection, req_list, dfs_stack):
...
-                    i = i + 1
-
-        rendering_information = None
-        response_body = None
-        if Found:
-            if valid_rendered_sequences_list[i].is_empty_sequence():
-                # then the current sequence will also be INVALID.
-                # propagate the root-cause explaining why the prefix was invalid
-                first_invalid_prefix = first_invalid_prefix_list[i]
-                logger.write_to_main(f"\tbut that prefix was INVALID (root = {first_invalid_prefix})\n")
```
PatGod, line 474: This code identifies the parent-most resource-type that prevents the execution of each target resource: that is why it is called the 'root', as in root-cause.

This information is printed in main.txt and is useful to identify bottlenecks in coverage.

All this information should be preserved, and not be removed.

restler/engine/core/driver.py
marina-p: The code to calculate sequences for directed smoke test is moved, not deleted.  See Driver.py Line ~507.  The main algorithm then only executes these sequences, including only once the prefix for each necessary sequence.

marina-p: The difference from the previous code is the order of the sequences: before, the algorithm was "render each prefix, up until the request being covered", now the algorithm is "for each generation, render all prefixes of a particular generation that will be required for the goal request."

PatGod: Note that the directed smoke test code already sorted the requests by their depth in the dependency graph. So the exploration was also generation by generation.

PatGod: But now, it seems root-cause identification is gone, cycle detection is gone, etc.

PatGod: Sorry, my mistake: cycle detection is still there, great!

Thank you also for the other clarifications.

PatGod (approved):

PR #225, opened 2021-05-25
Fix python grammar break due to string quoting

marina-p:
Example values that are already fully escaped should not be escaped again.


src/compiler/Restler.Compiler/CodeGenerator.fs
```diff
@@ -804,7 +804,12 @@ let getRequests(requests:Request list) includeOptionalParameters =
                 if s.Length > 1 then
                     s.Replace("'", "\\'"), "'"
                 else s, "'"
-            else if s.Contains("\"") then s.Replace("\"", "\\\""), "\""
+            // Special case already escaped quoted strings (this will be the case for example values).
```
PatGod, line 807: 'Special case *for* already...'?

PatGod (approved):

PR #228, opened 2021-05-27
Added logging for execptions to fix #220

### LINKED ISSUE #220: don't swallow errors in the engine
There are many places in the engine, especially on the fuzzing thread, where exceptions are swallowed.  We need to make sure exceptions are reported to be able to quickly diagnose errors.

This work item is opened to fix recently observed cases.  For example: 

        try:
            with open(grammar_path, 'r') as grammar:
                schema_json = json.load(grammar)
        except Exception as err:
            logger.write_to_main(f"Failed to process grammar file: {grammar_path}; {err!s}", print_to_console=True)
            sys.exit(-1)

    except ValueError as err:
        logger.write_to_main(f"Failed to parse grammar file for examples: {err!s}", print_to_console=True)
        return False

chakeson:
Added logger.exception() with message where it seemed fitting with issue #220.
It's worth noting when you use "except:" it catches all errors just not the once in [Exception](https://docs.python.org/3/library/exceptions.html). Most commonly you catch something like KeyboardInterupt.

## PR Checklist
* [x] Applies to work item: #220
* [x] Tests added/passed.  
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different approach. Issue number where discussion took place: #220

### DIFF
--- a/restler/engine/core/driver.py
+++ b/restler/engine/core/driver.py
@@ -144,6 +144,7 @@ def apply_checkers(checkers, renderings, global_lock):
                 RAW_LOGGING(f"Checker: {checker.__class__.__name__} kicks out\n")
         except Exception as error:
             print(f"Exception {error!s} applying checker {checker}")
+            logger.exception(f"Exception {error!s} applying checker {checker}")
             raise
 
 
@@ -664,6 +665,9 @@ def generate_sequences(fuzzing_requests, checkers, fuzzing_jobs=1):
                 logger.write_to_main("Exhausted collection...")
                 seq_collection = []
                 seq_collection_exhausted = True
+            
+            except Exception as ERR:
+                logger.exeception(f"Failed rendering in generate_sequences")
 
             logger.write_to_main(
                 f"{formatting.timestamp()}: Generation: {generation} / "
--- a/restler/restler.py
+++ b/restler/restler.py
@@ -61,6 +61,7 @@ def import_grammar(path):
         target_path = os.path.join(logger.EXPERIMENT_DIR, grammar_file)
         shutil.copyfile(path, target_path)
     except shutil.Error:
+        logger.exception("Failed to aquire grammar path.")
         pass
 
     return req_collection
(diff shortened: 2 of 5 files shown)


restler/engine/core/driver.py
```diff
@@ -664,6 +665,9 @@ def generate_sequences(fuzzing_requests, checkers, fuzzing_jobs=1):
                 logger.write_to_main("Exhausted collection...")
                 seq_collection = []
                 seq_collection_exhausted = True
+            
+            except Exception as ERR:
+                logger.exeception(f"Failed rendering in generate_sequences")
```
marina-p, line 670: >exeception [](http://example.com/codeflow?start=23&length=10)

exception

restler/restler.py
```diff
@@ -61,6 +61,7 @@ def import_grammar(path):
         target_path = os.path.join(logger.EXPERIMENT_DIR, grammar_file)
         shutil.copyfile(path, target_path)
     except shutil.Error:
+        logger.exception("Failed to aquire grammar path.")
```
marina-p, line 64: >aquire [](http://example.com/codeflow?start=36&length=6)

acquire

restler/restler.py
```diff
@@ -312,15 +315,18 @@ def signal_handler(sig, frame):
         settings = restler_settings.RestlerSettings(user_args)
     except restler_settings.InvalidValueError as error:
         print(f"\nArgument Error::\n\t{error!s}")
+        logger.exception(f"Argument error with {user_args}")
         sys.exit(-1)
     except Exception as error:
         print(f"\nFailed to parse user settings file: {error!s}")
+        logger.exception(f"Argument error with {user_args} failed to parse settings file.")
```
marina-p, line 322: >Argument error with {user_args} f [](http://example.com/codeflow?start=27&length=33)

Could you please change this to 'Failed to parse settings file with {user_args}'?

restler/restler.py
```diff
@@ -364,9 +372,11 @@ def signal_handler(sig, frame):
                 "update the request in the replay log with a valid authorization token.",
                 print_to_console=True
             )
+            logger.exception("Failed to play sequence from log.\n No valdi authorization token.")
```
marina-p, line 375: >valdi [](http://example.com/codeflow?start=69&length=5)

valid

marina-p:
Thanks a lot for implementing this!

I left a few comments.  

One more request: could you please also add a brief note to the description on how you tested this change?  e.g. "Tried RESTler with the following command line passing it an invalid grammar, and here is the output that I got"

marina-p (changes requested):
:clock1:

PR #239, opened 2021-06-04
fix JSON-to-Python null-to-None conversion when reading examples

### LINKED ISSUE #238: Payloadbody Checker mistakenly converts 'null' to 'None' values extracted from examples
This bug is due to a wrong JSON-to-Python conversion when reading values extracted from examples (via parsing grammar.json).

Self-assigned -- will fix shortly.

PatGod:
Closes #238 

Tests: manually with the PayloadBody Checker and a grammar.json including 'null' values


marina-p (approved):
:shipit:

PR #244, opened 2021-06-15
Track parameter combinations in the spec coverage file.

marina-p:
This change adds a new option to track parameter combinations to the engine.

When the fuzzing mode is 'test-all-combinations',
a new property 'tracked_parameters' will appear in speccov.json.
This property contains key-value pairs of all of the parameters
for which more than one combination is being tested.

In this commit, the parameters for which tracking is supported are:

- enums
- custom payloads

For example, an enum 'per_page' with several values will appear as:

        "tracked_parameters": {
            "per_page_9": "2"
        }

The suffix '9' is used to disambiguate between several primitives
of the same name appearing in the payload (for example "location" appears several times in a body), and is the position of the argument
in the request definition.

Full support for tracking fuzzable primitives will be enabled in a future
update.  For fuzzable primitives, the name of the parameter or property in the payload
is not yet included, and a default 'tracked_param' name is used instead.

For example:

        "tracked_parameters": {
            "tracked_param_14": "1"
        }

### DIFF
--- a/docs/user-guide/Testing.md
+++ b/docs/user-guide/Testing.md
@@ -132,6 +132,10 @@ the coverage data is being reported.
 * The __"error_message"__ value will be set to the response body if the request was not "valid".
 * The __"request_order"__ value is the 0 indexed order that the request was sent.
   * Requests sent during "preprocessing" or "postprocessing" will explicitely say so.
+* The __"tracked_parameters"__ property is optional and generated only when using
+`Test` mode with ```test-all-combinations```.
+This property contains key-value pairs of all of the parameters
+for which more than one value is being tested.
 
 #### Postprocessing Scripts:
 The `utilities` directory contains a sub-directory called `speccovparsing` that contains scripts for postprocessing speccov files.
--- a/docs/user-guide/Testing.md
+++ b/docs/user-guide/Testing.md
@@ -85,6 +85,7 @@ During each Test run a `speccov.json` file will be created in the logs directory
         "status_code": "400",
         "status_text": "BAD REQUEST",
         "error_message": "{\n    \"errors\": {\n        \"id\": \"'5882' is not of type 'integer'\"\n    },\n    \"message\": \"Input payload validation failed\"\n}\n",
+        "request_order": 4,
         "sample_request": {
             "request_sent_timestamp": null,
             "response_received_timestamp": "2021-03-31 18:20:14",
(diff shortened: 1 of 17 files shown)


PatGod:
Thanks!

Could you please update the section on 'speccov.json' (including the example) in Testing.md and add an explanation about these new 'tracked_parameters'?

PatGod (dismissed):

PatGod:
On second thought, since this change only affects the new option 'test-all-combinations', feel free NOT to update the general documentation... Still some form of documentation (other than the text of this PR) would be nice...

marina-p:
> On second thought, since this change only affects the new option 'test-all-combinations', feel free NOT to update the general documentation... Still some form of documentation (other than the text of this PR) would be nice...

I added the new data to the speccov example in Testing.md.

PatGod:
OK. Of course, the tracked_parameter field will not be included in the default speccov.json data unless the user explicitly uses test-all-combinations, right?

Shouldn't we add something like this in Testing.md?

"The tracked_parameter field in speccov.json is optional and generated only with the option test-all-combinations.
When the fuzzing mode is 'test-all-combinations',
a new property 'tracked_parameters' will appear in speccov.json.
This property contains key-value pairs of all of the parameters
for which more than one combination is being tested."

> Full support for tracking fuzzable primitives will be enabled in a future
update.

Unless you want to address this in that future...

PatGod (approved):

PatGod:
Oops, did you see my last suggestion above?

(You seem to have pushed changes at the same time.) Please ack but feel free to ignore - you decide. Thanks!

marina-p:
> Oops, did you see my last suggestion above?
> 
> (You seem to have pushed changes at the same time.) Please ack but feel free to ignore - you decide. Thanks!

I did not see it, that update to Testing.md sounds good.

marina-p:
p.s. This is not ready to be merged yet.  Somehow, this change has changed the behavior of the payload body checker.  The unit test is failing because there are now 4 bugs found instead of 1!  @PatGod since you just looked at that code recently, if there is a part of this change that you would suspect could cause the payload body checker to change the payloads it sends, please let me know.

PatGod:
@marina-p Regarding more bugs found by the payloadbody-checker, if you use the test-all-combinations option instead of the regular Test mode, the payloadbody-checker will likely kick in for every renderings of every request. So yes, it makes sense that, with the test-all-combinations option, more bugs may be found, but they are likely to be duplicates of each other.

If you meant that more bugs are found by the payload-body checker in regular Test mode, then it is suspicious indeed.

PatGod:
PS: note that yes, the payloadbody-checker is pseudo-random, so its behavior may be somewhat nondeterministic (for instance, of the search space is large and there are few bugs), tests checking the results of that checker (and others) may be 'flaky'. Did your recent experiments uncover a new source of flakiness in the unit-tests? (Or is this the known flakiness we already know about?)

marina-p:
I see, ok thanks for the info.  I have never seen flakiness from the test_payload_body_checker test before, and the results are deterministic (ran the test a few times both before and after, same result).  This is a unit test, so it is passing the same smoke test mode every time, no recent changes to the test itself.  Will investigate further...

PatGod (approved):

PR #253, opened 2021-06-30
Add the ability to track parameter values in the spec coverage file.

marina-p:
For large-scale testing, it is useful to have the specific values of
parameters fuzzed in the spec coverage file, in order to facilitate
root cause failure analysis.

This change adds an option to the compiler that produces a grammar which
contains the parameter names to track for every fuzzable value.

Testing:
- manual testing
- added/fixed engine tests to make sure new code path is exercised
- Note: the payload body checker code is not modified in this change because
it does not require this feature.


marina-p:
Add documentation to spec coverage format.

PatGod (dismissed):

PR #275, opened 2021-07-29
Add keyword for the body in the example config file.

marina-p:
Today, the schema parameter name for the body must be specified in the example config file (because parameters are required to match the schema at this time).  In order to easily try out different bodies without having to refer to the spec, this change adds a special keyword __body__ for specifying the body. 

This change also fixes a bug (and test bug) in inferring examples that was found while testing this change.

Also added diagnostic output on parameters in the examples that were not found in the spec.

### DIFF
--- a/docs/user-guide/Examples.md
+++ b/docs/user-guide/Examples.md
@@ -48,6 +48,12 @@ When specifying example payloads through an example file, set *DiscoverExamples*
 
 The format of the example file is as follows.  Note that either an external file or inline example can be specified.  Inline examples are useful for initial setup/debugging, but we recommend using external files to maintain examples separately for different requests.
 
+For specifying example body payloads, it is recommended to use the special keyword ```__body__```.
+This special property name will cause RESTler to assign this property value to the body,
+ignoring the name of the body parameter in the schema.  Because the body parameter name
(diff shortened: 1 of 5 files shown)


docs/user-guide/Examples.md
```diff
@@ -48,6 +48,9 @@ When specifying example payloads through an example file, set *DiscoverExamples*
 
 The format of the example file is as follows.  Note that either an external file or inline example can be specified.  Inline examples are useful for initial setup/debugging, but we recommend using external files to maintain examples separately for different requests.
 
+For specifying example body payloads, the special keyword ```__body__``` may be used
```
PatGod, line 51: Can you please perhaps add here what the effect of this special keyword is? For instance, does it override/skip schema validation when injecting the example body?

In other words, when should one use this new keyword and why?

Thanks!

PatGod (approved):

docs/user-guide/Examples.md
marina-p: Yes, thanks, I will make this language more direct (it is recommended to use this field when specifying example bodies).

PatGod: OK. Why is this recommended? (what is the effect?) Thanks!

PatGod (approved):

Issue #114, opened 2020-12-16
Support producers that return no content


JoshuaEN:
Hello,

I have an endpoint (in an existing API) that creates a resource, but returns 204 (No Content) and no Location header. The request itself contains the identifier for the created resource, and RESTler could (in theory) get the dependency information from the request body rather than the response.

I'm not sure how common this pattern is. Would supporting this use-case be in-scope for RESTler?

Thanks!

PatGod:
Hi Joshua, thanks for your suggestion. We'll investigate how this scenario could be supported... (Because of the upcoming holidays, please expect some delay...) 

Happy Holidays!

JoshuaEN:
Thanks! Happy Holidays!

PR #289, opened 2021-08-17
Enable producers from the input payload in the engine

marina-p:
Today, RESTler only supports producer-consumer dependencies by
parsing producers out of the response.

This enables a producer to be specified only as an input parameter.

In the grammar, a new named parameter called 'writer' is added to
the ```custom_payload_uuid4_suffix``` parameters.  The generated value
is then written similarly to how values would be parsed out of the response,
to be used in subsequent requests.

(Currently, only uuid4_suffix custom payloads are supported but we may
add support for custom payloads if that is a useful scenario.)

This partially implements #114

### DIFF
--- a/restler/engine/core/request_utilities.py
+++ b/restler/engine/core/request_utilities.py
@@ -143,7 +143,7 @@ def resolve_dynamic_primitives(values, candidate_values_pool):
 
     """
     global last_refresh, latest_token_value, latest_shadow_token_value
-
+    from utils import logger
     # There should only be one uuid4_suffix in the request for a given name
     current_uuid_suffixes = {}
     for i in range(len(values)):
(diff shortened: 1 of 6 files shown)


PatGod (approved):
Shouldn't this new feature/option be documented somewhere (Compiler settings?) in some .md file?

marina-p:
@PatGod This is just the engine part.  Since we don't document the grammar (unless I am forgetting something...let me know), I did not add any docs with this change.  I will add a new doc with the compiler change that will be checked in separately.

---
In reply to: [731262425](https://github.com/microsoft/restler-fuzzer/pull/289#issuecomment-731262425) [](http://example.com/codeflow?ancestors=731262425,731262425)

PR #294, opened 2021-08-19
Enable input parameters to be used in dependencies

marina-p:
This change adds support for producers that are not returned in the response.

Currently, this is supported via annotations only (which are specified in the same
way as usual).  For example, if PUT /A/{aId} does not return 'aId' in the
response, prior to this change it was not possible for GET /a/{aId} to use
the same ID if they are uniquely generated.  This is enabled by this change
by adding an annotation:

    {
      "producer_endpoint": "/A/{aId}",
      "producer_method": "PUT",
      "producer_resource_name": "aId",
      "consumer_param": "aId"
    }

In the future, scenarios such as the one above will also be enabled without
having to provide annotations.

See #114

### DIFF
--- a/docs/user-guide/Annotations.md
+++ b/docs/user-guide/Annotations.md
@@ -2,6 +2,23 @@
 
 RESTler allows the user to provide annotations with information about producer-consumer dependencies between requests.  This is helpful when RESTler cannot automatically infer such dependencies, which can happen for a variety of reasons (most often, because of very generic or inconsistent naming in the specification, or if the naming convention is not implemented in RESTler).
 
+## Supported types of dependencies
+RESTler currently supports the following types of producer-consumer dependencies:
+1. Between a response property and a request parameter.
+For example: request ```POST /A``` returns property "id", and request ```GET /A/{aId}``` specifies this ID in the path.
+
+2. Between two request parameters in different requests, where one of these is a ```custom_payload_uuid_suffix```.
+For example: ```POST /A/{aId}``` creates a resource with ID ```aId```, and ```GET /A/{AId}``` must use this ID, but the ID is not returned in the response.
+
+The annotation format for this annotation is the same as for annotations for producers that are response properties (as in (1)). RESTler will first look for the property name in the response, and if it does not find it will create this new kind of "input producer". The annotation does not explicitly allow to specify whether the ID should be extracted out of the response or not.
+
+3. Between two properties in the body.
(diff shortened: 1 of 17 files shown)


docs/user-guide/Annotations.md
```diff
@@ -2,6 +2,21 @@
 
 RESTler allows the user to provide annotations with information about producer-consumer dependencies between requests.  This is helpful when RESTler cannot automatically infer such dependencies, which can happen for a variety of reasons (most often, because of very generic or inconsistent naming in the specification, or if the naming convention is not implemented in RESTler).
 
+## Supported types of dependencies
+RESTler currently supports the following types of producer-consumer dependencies:
+1. Between a response property and a request parameter.
+For example: request ```POST /A``` returns property "id", and request ```GET /A/{aId}``` specifies this ID in the path.
+
+2. Between two request parameters in different requests, where one of these is a ```custom_payload_uuid_suffix```.
+For example: ```POST /A/{aId}``` creates a resource with ID ```aId```, and ```GET /A/{AId}``` must use this ID, but the ID is not returned in the response.
```
PatGod, line 11: in the POST response.

PatGod (approved):

PatGod:
This a great new feature!

However, I am confused: is there a new annotation that is required to enable this feature? If so, what is it, and why is it not explicitly added in the .md file? Please clarify. Thanks!

marina-p:
@PatGod Yes, it's currently required to add an annotation (as a starting point, this requires the least amount of testing to enable).

The annotation format is the same as if it were in the response.  RESTler will first look for the property name in the response, and if it does not find it will create this new kind of "input producer".  The annotation does not explicitly allow you to say if you want to extract the ID out of the response or not.   Does this clarify / if yes, should I add this text to the Annotation.md?  Or, what would you recommend to add?

---
In reply to: [901554282](https://github.com/microsoft/restler-fuzzer/pull/294#issuecomment-901554282) [](http://example.com/codeflow?ancestors=901554282,901554282)

PatGod:
Yes, thank you for your explanation, it answers my question. Yes, could you please perhaps add this part to the doc/.md file:

The annotation format is the same as if it were in the response. RESTler will first look for the property name in the response, and if it does not find it will create this new kind of "input producer". (Thus, the annotation does not explicitly allow to specify whether the ID should be extracted out of the response or not.)

Thanks much!

PR #303, opened 2021-08-24
Assign the correct primitive types in dependencies

marina-p:
The default type for a dependency was being assigned to a string.
However, it should be the original type of the property or parameter, since
this will determine how it is quoted.

(This bug was found when testing examples generated by the engine, directly
from the schema.)

While testing this issue, also fixed how array dependencies are handled, so
now both an array element and the array itself may be a producer.  Added a
unit test for this new behavior, which also confirms that the
dynamic object value is not quoted.

### DIFF
--- a/src/compiler/Restler.Compiler.Test/ExampleTests.fs
+++ b/src/compiler/Restler.Compiler.Test/ExampleTests.fs
@@ -74,6 +74,26 @@ module Examples =
                         ExampleConfigFilePath = Some exampleConfigFile }
 
         [<Fact>]
+        let ``array example where the array itself is a dynamic object`` () =
+            let grammarDirectoryPath = ctx.testRootDirPath
+            let config = { Restler.Config.SampleConfig with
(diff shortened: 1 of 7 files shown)


src/compiler/Restler.Compiler.Test/ExampleTests.fs
```diff
@@ -44,7 +44,7 @@ module Examples =
 
         [<Fact>]
         let ``array example in grammar without dependencies`` () =
-            let grammarDirectoryPath = ctx.testRootDirPath
+            let grammarDirectoryPath = @"c:\temp\restler" //ctx.testRootDirPath
```
PatGod, line 47: why is ctx.testRootDirPath now commented out? (is this left-over ad-hoc test code?)

marina-p: Thanks for catching, fixed.

PatGod (approved):

PR #309, opened 2021-08-25
Add dynamic objects to the engine schema parser.

marina-p:
Before, the schema parser simply used the dynamic object identifier
as a string (so any payload containing a dynamic object was invalid).

This change fixes this by getting the dynamic object value.

Testing: manual testing with real-world example.

Also contains a small change to improve exception handling in restler.py.


marina-p:
[This is the same change as PR #308 but merging to the correct branch.]

PR #319, opened 2021-08-30
Fix incorrect handling of HEAD method.

### LINKED ISSUE #317: RESTler times out with HEAD on Blob
RESTler is timing out with HEAD Api on blob. HEAD response headers contain Content-Length field which represents size of the blob. 
HEAD’s return body is empty.
RESTler thinks it has to read content-length bytes too and recvRepsonse  times out as there is no data corresponding to it. 

Below is an example request. 

2021-08-18 04:19:01.461: Sending: 'HEAD /xnsaccount1/containers71eb7015107b49258bbfc1f3dec4f129/blob77d6821ee1?timeout=30&se=2021-11-07T16:37:03.0591079Z&sv=2021-12-02&ss=b&sp=rwdylacup&srt=sco&spr=http,https&sig=W5Zd2In8d2lEJwpWISJwr9Rmdkz2wdWm2x%2BT2qH786Q%3D HTTP/1.1\r\nAccept: application/json\r\nHost: localhost:10000\r\nx-ms-test-restler-fuzzing-enabled: true\r\nx-ms-test-whitelist-api: false\r\nx-ms-client-request-id: fuzzstring\r\nContent-Length: 0\r\nUser-Agent: restler/8.0.0\r\n\r\n'

2021-08-18 04:21:01.500: Reached max req_timeout_sec of 120.

marina-p:
A response of the HEAD method does not have a body,
but returns the content-length of the body that would have been returned.
RESTler needs to handle this when reading the response.

Closes #317.

### DIFF
--- a/restler/engine/transport_layer/messaging.py
+++ b/restler/engine/transport_layer/messaging.py
@@ -72,6 +72,11 @@ class HttpSock(object):
         if self._sock:
             self._closeSocket()
 
+    def _get_method_from_message(self, message):
+        end_of_method_idx = message.find(" ")
+        method_name = message[0:end_of_method_idx]
+        return method_name
+
     def sendRecv(self, message, req_timeout_sec=600):
         """ Sends a specified request to the server and waits for a response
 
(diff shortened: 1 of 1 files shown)


PatGod (approved):
Thanks for this fix! (FYI, I have seen similar failures in recent ASF experiments, this fix should also improve coverage there too, great! ;-)

PR #320, opened 2021-08-31
Allow specifying max_combinations settings in test combination settings.

marina-p:
**BREAKING CHANGE** the settings for test combinations now have an object instead of a single value.

This additional setting allows configuring different RESTler runs with different maximum values for
combinations.  For example, shorter CI/CD invocations may be configured with fewer combinations than
longer or one-off regression tests.

### DIFF
--- a/docs/user-guide/SettingsFile.md
+++ b/docs/user-guide/SettingsFile.md
@@ -110,10 +110,13 @@ Testing
 different combinations of headers is supported via the following property:
 ```json
 "test_combinations_settings": {
-      "header_param_combinations" : "optional"
+     "header_param_combinations": {
+      "max_combinations": 50,
+      "param_kind": "optional"
+    }
 }
 ```
-The supported values are 'optional', 'required', and 'all'.
+The supported ```param_kind``` values are 'optional', 'required', and 'all'.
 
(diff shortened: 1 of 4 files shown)


PatGod (approved):

mbiuki:
Are changes supersede the global `max_combinations`, for example if I have:

```json
	"max_combinations": 100,    <------ Global max_combinations
	"dyn_objects_cache_size": 40,
	"test_combinations_settings": {
		"header_param_combinations": {
			"max_combinations": 10,   <---- max_combinations of header params
			"param_kind": "optional"
		},
		"query_param_combinations": {
			"max_combinations": 15, <---- max_combinations of body params
			"param_kind": "required"
		},
		"example_payloadds": {
			"payload_kind": "all"
		}
	},
```
Which `max_combinations` are in prior?

marina-p:
Hi @mbiuki,

For parameter combinations: the local ```max_combinations``` take precedence over the global value.  So, this should be 10 in the above example.

However, for overall combinations (schema + values), the usual global setting of 'max_combinations' is used.

In the above example, 100 combinations may still be tested if there are enough value combinations.  If there are more than 100 combinations of values for one of the parameters, then multiple parameter combinations will not be tested.

PR #325, opened 2021-09-01
Easily replace the entire body of a request.

marina-p:
This change adds the ability to specify that the entire body of a request type should be
replaced by custom values specified in the dictionary.

### DIFF
--- a/docs/user-guide/FuzzingDictionary.md
+++ b/docs/user-guide/FuzzingDictionary.md
@@ -29,6 +29,19 @@ The following describes how each property in the dictionary is used in a RESTler
   }
   ```
 
+A special syntax may be used to replace the entire contents
+of the body of a request.  For a request with endpoint
+```/api/blog/{blogId}``` and method ```GET```, the following replaces the body of this request with the
+specified custom payloads.
+
+  ``` json
+  {
+     "restler_custom_payload": {
+         "/api/blog/{blogId}/get/__body__": "new body"
+     }
+  }
+  ```
+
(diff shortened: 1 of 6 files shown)


PatGod (approved):
Isn't this new feature redundant with the ability to provide request examples?

Or is this new feature when a full example is not available, but only a body?

Also, when using this feature, will the provided body be used without any modification (as a straight 'static_string'), or might some specific values substituted according to other instructions in the fuzzing dictionary (like restler_custom_payload etc.)?

Would you mind perhaps clarifying this in FuzzingDictionary.md?

marina-p:
Yep, it is meant to simply replace the body as a constant string (without any further substitutions).

marina-p:
The reason it's being added is to address the lack of ability to specify per-resource custom payloads in the dictionary.  Replacing a specific body parameter does not work well with the current dictionary because every body parameter in the spec could be named "body".  Another alternative would be to add per-resource parts of the dictionary and update the engine to consume them (we discussed a while ago and decided not to add this to avoid increasing the complexity of the dictionary).  Or, the syntax of this change could be extended to work for any property, not just the body.


```json
  "per_resource_custom_payloads": {
    "/api/blog/{blogId}": {
      "put": {
        "restler_custom_payload" :{
           "id": 123
        }
      },
      "post": {
        "restler_custom_payload" :{
          "id": 456,
          "message": "new"
       }
      }
    }
  }
```

PatGod:
Thank you for the detailed explanation. Perhaps we should keep your suggestion above in mind, see if users ask for more than what is provided in this PR, and then revisit then if needed.

PR #358, opened 2021-10-04
Support multiple authentication headers in requests

victor-li:

Currently, scripts to generate authentication tokens can only provide one single authentication header for API requests.
For various use cases (such as API sandboxes), multiple authentication headers are required to allow the API to function properly (e.g. additional API keys).

This PR allows authentication scripts to provide multiple authentication headers for all requests, where headers are separated with a semicolon.

The output of the authentication script could the be:
```
{u'app1': {<any additional metadata you'd like to print. currently only used for troubleshooting. >}, u'app2':{}}
ApiTokenTag: 9A;ApiTokenTag2: 85
ApiTokenTag: ZQ;ApiTokenTag2: UE
```

## PR Checklist
* [ ] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign.
* [ ] Tests added/passed.  
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different approach. Issue number where discussion took place: #xxx


marina-p:
Hello Victor,

Thank you for contributing this new use case for Authentication.

One request, could you please make it so that the tokens are one per line (to avoid issues with using a specific delimiter)?

The format could be:

```
{ ... }
ApiTokenTag: 9A
ApiTokenTag2: 9B
---
ApiTokenTag: ZQ
ApiTokenTag2: BZZ
```

To get the tokens for all the users, the code can first check for the '---' delimiter on a separate line, and if it is found use the new 'multiple tokens per user' code path, otherwise use the current logic of one token per line corresponding to each user.

marina-p (changes requested):
:clock1:

victor-li:
Hi Marina,

I agree that the delimiter I proposed could lead to potential issues, since semicolons are valid in header values. However, your proposal can lead to the following confusing situation:
```
{u'app1': {}}
ApiTokenTag: 9A
ApiTokenTag2: ZQ
```

```
{u'app1': {}, u'app2':{}}
ApiTokenTag: 9A
ApiTokenTag: ZQ
```

In the first case, there is only one application requiring two tokens (delimited by `---`), while in the second case there are two applications requiring one token each.

We could make the delimiter mandatory for multiple applications, which results in:
```
{u'app1': {}, u'app2':{}}
ApiTokenTag: 9A
---
ApiTokenTag: ZQ
```

What do you think?

marina-p:
Hi Victor, I was thinking the above case would be covered in the special case, as below.  Let me know if I am still missing something.  


```python
import sys
import ast
with open("c:\\temp\\token.out", 'r') as fr:
   str = fr.read()
   tokens = str.split("---")

   metadata = tokens[0].split("\n")[0]
   metadata = ast.literal_eval(metadata)

   if len(tokens) == 1 or len(metadata) > len(tokens):
      # Code path for single token per app, parse each line
      app_tokens = tokens[0].split("\n")
      app_tokens = app_tokens[1:]

      for idx, app in enumerate(metadata):
         print(f"App: {app}, Token list: {app_tokens[idx]}")

   else:
      for idx, app in enumerate(metadata):
         app_tokens = tokens[idx]

         app_tokens = list(filter(bool, app_tokens.split("\n")))
         if idx == 0:
            app_tokens = app_tokens[1:]

         print(f"App: {app}, Token list: {app_tokens}")
```

---
In reply to: [934339575](https://github.com/microsoft/restler-fuzzer/pull/358#issuecomment-934339575) [](http://example.com/codeflow?ancestors=934339575,934339575)

victor-li:
I understand that the logic can determine whether the tokens should be parsed individually on each line or look for the delimiter for multiple applications. The only concern I had was that the examples I provided look very similar, while the tokens are parsed in a very different way. Nevertheless, I am fine with your proposal, which also provides backwards compatibility.

PatGod:
Thanks @victor-li . Please note that backwards-compatibility is (of course) super important here! We cannot accept any solution that is not backwards-compatible.

marina-p:
Hi Victor, 
Ah, ok I understand the concern now, thank you for clarifying.

Another option is that RESTler could support both a new, more structured json format that would handle better this confusing case, and the current txt format that would only be supported for the 'one token per app' scenario.  This would be a bit more work to implement, so in my opinion it would be fine to defer it until more use cases need to be added with additional data for each app.

restler/engine/core/request_utilities.py
```diff
@@ -104,6 +96,43 @@ def execute_token_refresh_cmd(cmd):
         _RAW_LOGGING(f"\nMaximum number of retries ({MAX_RETRIES}) exceeded. Exiting program.")
         sys.exit(-1)
 
+def parse_authentication_tokens(cmd_result):
+    """ Parses the output @param cmd_result from token scripts to refresh tokens.
+
+    @param cmd_result: The user-provided command to refresh the token.
```
marina-p, line 102: >The user-provide [](http://example.com/codeflow?start=23&length=16)

The reslut of running the user-provided

restler/unit_tests/test_auth_parser.py
```diff
@@ -0,0 +1,24 @@
+import unittest
+
+from engine.core.request_utilities import parse_authentication_tokens, NO_TOKEN_SPECIFIED, NO_SHADOW_TOKEN_SPECIFIED
+
+class AuthParserTest(unittest.TestCase):
```
marina-p, line 5: >AuthParserTest [](http://example.com/codeflow?start=6&length=14)

Thanks for adding this!

marina-p: FYI, we do not have any end to end coverage for authentication in our CI pipeline.  It would be great if you can add a note to the PR that you also manually tested this on your real-world API example and it worked as expected.

marina-p:
Looks great.  I made a few minor comments.  Also, please update the PR description to the final design.

---
In reply to: [942163876](https://github.com/microsoft/restler-fuzzer/pull/358#issuecomment-942163876) [](http://example.com/codeflow?ancestors=942163876,942163876)

marina-p (approved):
:shipit:

restler/engine/core/request_utilities.py
```diff
@@ -104,6 +96,43 @@ def execute_token_refresh_cmd(cmd):
         _RAW_LOGGING(f"\nMaximum number of retries ({MAX_RETRIES}) exceeded. Exiting program.")
         sys.exit(-1)
 
+def parse_authentication_tokens(cmd_result):
+    """ Parses the output @param cmd_result from token scripts to refresh tokens.
```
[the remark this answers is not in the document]
victor-li: > If I understand correctly, they are delimited by '\r\n' and the rest of the code simply plugs this into the list of headers, which is why no other code changes were needed.

That is correct.

Issue #348, opened 2021-09-16
Why are there more values fuzzed than specified in the grammar? 


c7h:
Hi!
I defined in my (custom) dict.json that a fuzzable datetime object  should look like this: 

```
  "restler_fuzzable_datetime": [
    "2019-06-26T20:20:39+00:00"
  ]
```
However, when I run _restler test_, I see in the network testing log that It's testing with a group like this:
```   
+ restler_fuzzable_datetime: ['"2019-06-26T20:20:39+00:00"', '"2021-10-17"', '"2021-09-15"']
                                                              ^---------------------------^
                                                                definitely not mine 
```
This leads to the tests failing since the format is not supported. How do I make sure only my dict is used?

mbiuki:
I've had the same problem reported here: https://github.com/microsoft/restler-fuzzer/issues/153#issuecomment-817670148 and here: https://github.com/microsoft/restler-fuzzer/issues/153#issuecomment-820941267 and their subsequent correspondence in that ticket #153 . 
Let me know if you get it to work and I would follow your sequence, otherwise, I still have the same issue. Thanks.

PatGod:
After modifying your dictionary (like done above), please re-compile and then re-test -- after re-compilation, the old 'definitely not mine' values should disappear. (If not, please let us know - that would be a bug.)

Thanks!

PatGod:
FYI, the need for re-compilation after modifying the fuzzing dictionary has been clarified (we hope!) recently in https://github.com/microsoft/restler-fuzzer/blob/main/docs/user-guide/FuzzingDictionary.md#how-to-configure-the-dictionary

If this is not clear, please let us know. Thanks again for your feedback.

mbiuki:
@PatGod , Thanks for your clarification. Maybe this is the solution to the problem that we are experiencing. I can follow your steps and let you know if this solved the issue. Nevertheless, manually copying config and dict out of the Compile folder seems a bit ad-hoc. Is there any other better way to address this challenge like the ideas that are passed around in #211, or even, caching and comparison when re-compile could do?

PatGod:
@mbiuki Well, in my opinion, the solution is always to compile with a config.json file, and never use the compile --api_spec option. 

The compile --api_spec option was created just before open-sourcing RESTler in an attempt to simplify the workflow for first-time users. 

[...]

PR #364, opened 2021-10-06
Fix issue #348

marina-p:
Do not add dates by default; add an engine setting that may be specified to keep the current behavior.

(Note: a further fix to the current behavior is needed to generate dates in the format corresponding to the
ones specified in the spec or dictionary.)


marina-p:
Fixes #348

PatGod (approved):

PR #381, opened 2021-10-19
Enable ordering constraints in the engine

marina-p:
When specified in the 'pre_send' section of the grammar, reader variables
are added to the _consumer list, in the same way as when they are
part of the payload.

Testing: tested manually by adding the following annotations to demo_server,
and confirming that every sequence starts with the GET request.
```json
{"x-restler-global-annotations":
[{
  "producer_endpoint": "/blog/posts",
  "producer_method": "GET",
  "consumer_endpoint": "/blog/posts",
  "consumer_method": "POST"
}]
}
```

### DIFF
--- a/restler/engine/core/requests.py
+++ b/restler/engine/core/requests.py
@@ -263,6 +263,12 @@ class Request(object):
             var_name = self._get_var_name_from_definition_line(line)
             if var_name:
                 self._consumes.add(var_name)
+        # Also look for reader placeholders in the pre_send section
+        if bool(self.metadata) and 'pre_send' in self.metadata\
+        and 'dependencies' in self.metadata['pre_send']:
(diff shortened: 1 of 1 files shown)


restler/engine/primitives.py
```diff
@@ -279,6 +279,8 @@ def _flatten_and_quote_candidate_values(candidate_vals):
                 if tag in candidate_values[primitive_name]:
                    return _flatten_and_quote_candidate_values(candidate_values[primitive_name][tag])
                 # tag not specified in per_endpoint values, try sending from default list
+                print(f"Tag: {tag}, prim: {primitive_name}")
```
marina-p, line 282: >Tag [](http://example.com/codeflow?start=24&length=3)

revert!

PR #382, opened 2021-10-20
Always try to parse responses from GET after async polling

marina-p:
Previously, if async polling from a PUT response in a certain format was received, that was used as the final
response from which to parse dynamic objects, otherwise the GET request was used.   This did not always work -
in some cases, there were properties missing from the response or an empty body was returned.

After this change, the first response on which the response parser will be invoked will be the one from the GET
request.  If the GET failed or parsing the response of the GET, the response from the PUT will still be tried as before.

Testing: ad-hoc testing using original repro (in progress).

### DIFF
--- a/restler/engine/core/async_request_utilities.py
+++ b/restler/engine/core/async_request_utilities.py
@@ -141,9 +141,13 @@ def try_async_poll(request_data, response, max_async_wait_time):
     @type  max_async_wait_time: Int
 
     @return: A tuple containing:
-             - The response for parsing, which will either be the @response
-             passed to this function or the response from the GET request
-             if it was an async request.
+             - The list of responses for parsing, which will either contain the @response
+             passed to this function, or in the case of an async request, the
+             response from the end of the async operation (if finished),
+             and the response from the GET request.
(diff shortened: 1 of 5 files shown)


PatGod (approved):

restler/engine/core/async_request_utilities.py
```diff
@@ -228,6 +244,8 @@ def try_async_poll(request_data, response, max_async_wait_time):
                 if get_response:
                     # Use response from the GET request for parsing. If the GET request failed,
                     # the caller will know to try and use the original PUT response for parsing
-                    response_to_parse = get_response
+                    response_to_parse.append(get_response)
```
PatGod, line 247: shouldn't this be 'responseS_to_parse.append(get_response) ?

PatGod: yes, this is a bug (in an error leg not exercised in manual testing -- JSONDecodeError) -- please fix this typo before merging.

PatGod: please fix the above

PatGod (changes requested):

PR #392, opened 2021-10-28
ci: full multi-stage docker build

rhtenhove:
Wanting to be able to run the latest main in Docker, I needed a Dockerfile which would do all the building for me.

The current Dockerfile had some quirks and could use a touch up, so I've rewritten it to do the entire build process, so the current state of the repo can quickly be built locally or in CI.

This uses multistage, so any source files will not clutter the final result. As such the final image is as small as possible.

I'm not sure how this will interfere with your current build and test pipelines, hopefully this can easily be integrated.

## latest

It would also be highly useful to have a `:latest` tag for this repo which would match a build from main. Could you implement that in the pipeline?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,17 +1,20 @@
-FROM mcr.microsoft.com/dotnet/aspnet:5.0-alpine
+FROM mcr.microsoft.com/dotnet/sdk:5.0-alpine as builder
 
-COPY ./compiler /RESTler/compiler
-COPY ./restler /RESTler/restler
+RUN apk add --no-cache python3 py3-pip
+RUN ln -s /usr/bin/python3 /usr/bin/python
 
-#Get the source code for the engine, and
-#copy source code to the image
-#RESTler Python dependency
-RUN apk add --no-cache python3
-RUN python3 -m ensurepip
-RUN pip3 install --upgrade pip
-RUN pip3 install requests
-RUN pip3 install requests
-RUN pip3 install applicationinsights
-COPY ./engine /RESTler/engine
-RUN python3 -m compileall -b /RESTler/engine
-COPY ./resultsAnalyzer /RESTler/resultsAnalyzer
+COPY src ./src
+COPY restler ./restler
+COPY build-restler.py .
+
+RUN python3 build-restler.py --dest_dir /build
+
+RUN python3 -m compileall -b /build/engine
+
+FROM mcr.microsoft.com/dotnet/aspnet:5.0-alpine as target
+
+RUN apk add --no-cache python3 py3-pip
+RUN ln -s /usr/bin/python3 /usr/bin/python
+RUN pip3 install requests applicationinsights
+
+COPY --from=builder /build /RESTler
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,20 +1,17 @@
-FROM mcr.microsoft.com/dotnet/sdk:5.0-alpine as builder
+FROM mcr.microsoft.com/dotnet/aspnet:5.0-alpine
 
-RUN apk add --no-cache python3 py3-pip
-RUN ln -s /usr/bin/python3 /usr/bin/python
+COPY ./compiler /RESTler/compiler
+COPY ./restler /RESTler/restler
 
-COPY src ./src
-COPY restler ./restler
-COPY build-restler.py .
-
-RUN python3 build-restler.py --dest_dir /build
-
-RUN python3 -m compileall -b /build/engine
-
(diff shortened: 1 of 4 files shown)


marina-p:
Hello @rhtenhove,

Got a chance to test this out w/ our pipeline (thanks for the ping).  I'll add some comments shortly.

Thanks,

Marina

docker/Dockerfile
```diff
@@ -1,17 +1,20 @@
...
-RUN pip3 install applicationinsights
-COPY ./engine /RESTler/engine
-RUN python3 -m compileall -b /RESTler/engine
-COPY ./resultsAnalyzer /RESTler/resultsAnalyzer
+COPY src ./src
+COPY restler ./restler
+COPY build-restler.py .
+
+RUN python3 build-restler.py --dest_dir /build
+
```
marina-p, line 11: The container build should fail if the build command fails - in my testing, it did not.

docker/Dockerfile
```diff
@@ -1,17 +1,20 @@
...
-RUN python3 -m ensurepip
-RUN pip3 install --upgrade pip
-RUN pip3 install requests
-RUN pip3 install requests
-RUN pip3 install applicationinsights
-COPY ./engine /RESTler/engine
-RUN python3 -m compileall -b /RESTler/engine
-COPY ./resultsAnalyzer /RESTler/resultsAnalyzer
+COPY src ./src
+COPY restler ./restler
```
marina-p, line 7: This indeed does not work with the existing pipeline, the context needs to be updated (I will point you to a version to use).

docker/Dockerfile
```diff
@@ -1,17 +1,20 @@
-FROM mcr.microsoft.com/dotnet/aspnet:5.0-alpine
+FROM mcr.microsoft.com/dotnet/sdk:5.0-alpine as builder
```
marina-p, line 2: General comment: could you please add a build argument to still allow the current use case of using a pre-built directory to build the container?  (could be "selfContained" for the new use case being introduced).  In some cases, we need to build, then run tests (not container based today), then copy the exact binaries that were tested on the container.

marina-p (changes requested):
:clock1:

docker/Dockerfile
rhtenhove: that is because of an odd method of catching build failures. I've added a simple replacement which uses the builder exit code to determine failure

docker/Dockerfile
rhtenhove: It isn't possible to do a conditional COPY (at least not without hacky tricks).

So what you need is to build and test binaries outside of a container, and then COPY those into an image, as currently on main. What will that image then be used for?

This proposal is to have a Dockerfile which can be built by anyone, anywhere, without any dependency on a specific environment. This would then also be the published image, so this is reproducible.

You can of course create a separate Dockerfile for the process you describe.

docker/Dockerfile
rhtenhove: The context is the root of this repo. So the build command would be `docker build -f docker/Dockerfile .` I've added this to the README.

It would probably be even more clear if the `Dockerfile` was moved to the root of the repo. Would you agree?

marina-p: Hello,

My apologies, just getting back to this now.  Yes, definitely, it would be more clear if the Dockerfile is at the root.  

How about we do two updates, so we do not need to coordinate the CI changes?

1) For this PR, add the Dockerfile with your latest changes to the root, and leave the docker/Dockerfile as is.  

2) I will follow up with a change to get CI working using the new dockerfile, and remove the old one.

docker/Dockerfile
marina-p: Looks good, thanks!

docker/Dockerfile
marina-p: Ok, if this is messy and non-standard, then it makes sense to have the standard Dockerfile be the default, which should be the version with your changes.  I'll try the approach of copying the binaries out of the container to sanity test them in CI before publishing (the image built in our pipeline is the one pushed to the official registry on DockerHub ([https://hub.docker.com/_/microsoft-restlerfuzzer-restler](https://hub.docker.com/_/microsoft-restlerfuzzer-restler)).

docker/Dockerfile
rhtenhove: Yes, agreed. I've changed the files as you've described in step 1.

docker/Dockerfile
rhtenhove: Excellent, sounds good.

marina-p (approved):
:shipit:

PR #418, opened 2021-12-03
Add the ability to specify a local logs directory.

### LINKED ISSUE #406: Provide option to output task logs into a specific directory
This should be specified on the command line, e.g.:

```restler.exe --output_log_dir  <directory>```

marina-p:
Closes #406


PatGod:
How does this differ from the existing option ---logsUploadRootDirPath (which is also not documented anywhere in the .md files)?

marina-p:
The ```--logsUploadRootDirPath``` option is for optionally copying logs from the local directory to some other, remote directory.  This new option is for changing what the local logs directory is.

There is no .md file for the command-line options today.  I'll open a separate task for this.

PR #423, opened 2021-12-06
Dockerfile for demo server

rhtenhove:
Wanting to test out RESTler with a simple demo server, I created a Dockerfile so this demo server can be easily run.

Since `FLASK_RUN_HOST` didn't work, I've added it to the run command. It will keep the default `localhost` binding, but allows setting it as `0.0.0.0` for when run in a container.

### DIFF
--- a/demo_server/Dockerfile
+++ b/demo_server/Dockerfile
@@ -1,0 +1,14 @@
+FROM python:3.8-alpine
+
+RUN apk add --no-cache build-base
+
+COPY demo_server/  /app/demo_server
+COPY requirements.txt logging.conf swagger.json /app/
+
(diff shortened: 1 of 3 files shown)


marina-p (approved):
:shipit:

marina-p:
Thanks for adding this!

marina-p:
Just FYI, there was a CI failure which I re-ran and it passed - looks like a rare flakiness issue.  (Opened #424 to track it.)

---
In reply to: [987401537](https://github.com/microsoft/restler-fuzzer/pull/423#issuecomment-987401537) [](http://example.com/codeflow?ancestors=987401537)

Issue #429, opened 2021-12-14
Payloadbody checker: refactor/strengthen/correct the uuid_substitution code


PatGod:
In the payloadbody checker, the following code should be improved following the suggestions indicated by "Bug:" below:

```
                    # Bug: tries all the uuid suffixes in the dictionary
                    # even though this payload may not contain most of them
                    for uuid4_suffix in uuid4_suffix_dict:
                        suffix = uuid4_suffix_dict[uuid4_suffix]
                        len_suffix = len(suffix)

                        # If the suffix is present in the path
                        # Bug: the code below assumes the path ends with the suffix
                        # and the suffix is unique, i.e. does not occur twice in the path;
                        # this cannot be assumed, since the user can make the suffix any value
                        if suffix in rendered_data[:partition]:
                            new_val_start = rendered_data[:partition].index(suffix)
                            if new_val_start + len_suffix + 10 > partition:
                                self._log('unexpected uuid')
                                continue
                            new_val = rendered_data[new_val_start:new_val_start + len_suffix + 10]

                            new_body = rendered_data[start_body:]
                            # find all occurence in the body
                            suffix_in_body = [
                                # Bug: Finds all occurrences of *the value text string* of suffix in the
                                # payload and does replacement
                                # Instead, should search in the grammar for the 'restler_custom_payload_uuid_suffix' and
                                # replace with new rendered value
                                # Note: the replacement done below can be incorrect in 2 ways in an example like this:
                                #   Ex: "name":"Standard" will be replaced by "nameaa1e45cbb5d"
                                # 1. the right-hand-side should be replaced, not the left-hand-side in a case like this
                                # 2. the length of old_val should be the length of suffix, not the length of new_val
                                m.start() for m in re.finditer(suffix, new_body)
                            ]
                            for si in suffix_in_body:
                                old_val = new_body[si: si + len_suffix + 10]
                                new_body = new_body.replace(old_val, new_val)
                            # replace the old body with the new_body
                            rendered_data = rendered_data[:start_body] + new_body
```

PatGod:
The newly-introduced setting skip_uuid_substitution might be removed when this work-item is done. (TBD)

PR #430, opened 2021-12-14
Bug fix: the payloadbody checker should not modify auth-tokens

### LINKED ISSUE #324: auth tokens are sometimes printed in the network.testing logs, even with no_tokens_in_logs on
Sometimes, the RESTler logs include auth tokens, despite the option no_tokens_in_logs being turned on.

This issue is to track progress on this.

PatGod:
Closes #324 

This PR fixes a bug in the payloadbody checker: it should not modify auth-tokens when substituting uuid-suffixes in the body of a request. (The bug was: auth-tokens were erroneously included in the 'body' part of a request.)

This PR also includes a new option skip_uuid_substitution to skip this part of the payloadbody checker. This new option can be turned on by adding the following in the engine settings.json file:

```
"checkers": {
    "payloadbody": {
        "skip_uuid_substitution" : true
    }
  },
```

Indeed, the uuid_substitution code should be rewritten - a new issue #429 was opened to track this work-item.

Testing: manually with a RESTler job where the payloadbody checker erroneously modified auth-tokens, resulting in tokens being printed in the network.testing logs and REST API requests returning "401 unauthorized" errors.

### DIFF
--- a/restler/checkers/payload_body_checker.py
+++ b/restler/checkers/payload_body_checker.py
@@ -65,6 +65,7 @@ class PayloadBodyChecker(CheckerBase):
         self._size_dep_budget = True
         self._setup_done = False
         self._use_feedback = True
+        self._skip_uuid_substitution = False
         self._current_task_tag = ''
         self._fuzzed_requests = set()
         self._buckets = PayloadBodyBuckets()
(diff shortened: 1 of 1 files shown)


restler/checkers/payload_body_checker.py
```diff
@@ -1135,31 +1138,59 @@ def _exec_request_with_new_body(
...
             rendered_data = seq.resolve_dependencies(rendered_data)
 
-            # substitute if there is UUID suffix
-            original_rendered_data = rendered_data
-            uuid4_suffix_dict = self._get_custom_payload_uuid4_suffix()
-            for uuid4_suffix in uuid4_suffix_dict:
-                suffix = uuid4_suffix_dict[uuid4_suffix]
-                len_suffix = len(suffix)
-                # need the query to partition path and body
+            if not self._skip_uuid_substitution:
```
marina-p, line 1141: >_skip_uuid_substitution [](http://example.com/codeflow?start=24&length=23)

Could you please add a comment referencing that this setting may be removed when issue #429 is fixed?

marina-p (approved):
:shipit:

restler/checkers/payload_body_checker.py
PatGod: @marina-p  Well, removing this option later won't be backwards compatible... Is this wise?

(The option is harmless and is off by default.) 

OK?

PatGod: I just added a note with issue #429 so we remember this point.

Issue #408, opened 2021-11-17
"IsRequired" wrong for InternalNode


JustinnT:
Good Day, thank you for this amazing tool.

I am currently testing out restler-compile on [this file](https://github.com/jdegre/5GC_APIs/blob/master/TS29510_Nnrf_NFDiscovery.yaml), TS29510_Nnrf_NFDiscovery.yaml.

Restler is able to compile successfully. However when I check the generated grammar.json, for some optional parameters, the "isRequired" option is wrongly set inside InternalNode. 

For example, in the grammar.json the optional parameter: "service-names" is set as "isRequired": true [here](https://github.com/JustinnT/restler_issue/blob/master/Compile/grammar.json#L167-L255). But from the [openAPI description](https://github.com/jdegre/5GC_APIs/blob/e14e017101b2d25f6e1bdb7af851d709c89eadbe/TS29510_Nnrf_NFDiscovery.yaml#L57-L67) of this optional parameter "service-names", it did not state that it is required, meaning that this should be not required/optional.

If you require for testing purposes, here is a [repo](https://github.com/JustinnT/restler_issue/) which includes the file in question and dependencies required for restler to compile, as well as my results from running.

Is this also related to the #399 ?

marina-p:
Hello @JustinnT,

We will look into this, thank you for reporting it.  This is not related to #399.   

Thanks,

Marina

JustinnT:
Hi @marina-p 

Thank you for the reply. For now, is there any workaround to prevent restler from including these parameters in the requests? 

I tried manually setting the grammar.json "isRequired": true -> "isRequired": false but it does not have an effect. 

Thanks,
Justin

marina-p:
Related issue: #408

PR #432, opened 2021-12-17
Make 'isRequired' in grammar.json match the spec

marina-p:
Prior to this change, the 'isRequired' flag was always set to 'true' for internal nodes
(arrays and objects with properties).  With the upcoming change to test
combinations of nested parameters, make these match the spec in grammar.json.

Note: there are no changes to grammar.py as part of this change - it should
still include all parameters.

Added baseline test for grammar.json and grammar.py.


marina-p:
Related issue: #408

PR #436, opened 2022-01-06
Added support for swagger escape characters

### LINKED ISSUE #414: Compiler does not respect $ref escape sequences.
OpenAPI supports $ref tags for internal and external file references. Sometimes, the value contains escaped sequences. OpenAPI3 specifies 2 escape sequences `~1` and `~0` (see https://swagger.io/docs/specification/using-ref/#escape). 

While compiling a specification which contains a $ref to _#/paths/~1nf-instances/get/parameters/25/schema_, RESTler fails with an error
```
Unhandled exception. System.AggregateException: One or more errors occurred. (Could not resolve the path '#/paths/~1nf-instances/get/parameters/25/schema'.)
 ---> System.InvalidOperationException: Could not resolve the path '#/paths/~1nf-instances/get/parameters/25/schema'.
   at NJsonSchema.JsonReferenceResolver.ResolveDocumentReference(Object rootObject, String jsonPath)
   at NJsonSchema.JsonReferenceResolver.ResolveReferenceAsync(Object rootObject, String jsonPath, Boolean append)
```

This might be caused by a bug in a third party library, however I still want to note that this also affects RESTler.

Best,
Christoph

Nauscar:
Support for escape characters in $ref as listed in the swagger specification: https://swagger.io/docs/specification/using-ref/
Closes #414


marina-p:
Hi @Nauscar, I played around with the original repro which contains a pointer to a schema in the path.  That one is not addressed by this fix.  I am wondering if we inline all such references whether it would start working.  Let me know if you have time to try out this path, or I can try this sometime tomorrow.

marina-p:
This is the snippet I created that I believe reproduces the issue originally reported:

```
  "paths": {
    "/escape3": {
      "post": {
        "operationId": "post_escape3",
        "parameters": [
          {
              "$ref": "#/paths/~1escape4/get/parameters/0"
          }
        ],
        "responses": {
          "200": {
            "description": "Success"
          }
        }
      }
    },
    "/escape4": {
      "get": {
        "parameters": [
          {
            "in": "query",
            "name": "query_param",
            "type": "string"
          }
        ],        "responses": {
          "200": {
            "description": "Success"
          }
        }

      }
    }
  },

```


---
In reply to: [1009364753](https://github.com/microsoft/restler-fuzzer/pull/436#issuecomment-1009364753) [](http://example.com/codeflow?ancestors=1009364753)

marina-p (dismissed):
:shipit:

marina-p (approved):
:shipit:

Issue #387, opened 2021-10-21
allow specifying base path in the engine settings


marina-p:
This is required for test environments, which may use different base paths than the one defined in the OpenAPI specification.

jason-invision:
We have a similar issue where the OpenAP spec defines a service contract that may be implemented by difference services and exposed via a different base path.

marina-p:
For how to implement this: 
A new solution is needed for how to identify the base path in the grammar (currently, it is simply appended to the front of the path).

Option 1: create ```restler_static_basepath``` primitive 
Option 2: generate a new custom payload named ```__basepath__```, and add this to the dictionary with the base path specified in the OpenAPI spec.    If there is no base path specified, the custom payload still needs to be generated, with one empty string value (so users could later add it to the engine settings).  The engine could then treat this custom payload as a special case, checking if it is specified in the engine settings, and otherwise using the static value specified in the dictionary.

For option 2, in Compiler.fs Line 745 - the basePath should be moved from the statement below into a separate one right above, which creates a FuzzingPayload.Custom for it (currently it turns into a FuzzingPayload.Constant in the else branch).  The corresponding custom payload entry also needs to be added into the dictionary (```newDictionary```).

```
    let path =
        (basePath + requestId.endpoint).Split([|'/'|], StringSplitOptions.RemoveEmptyEntries)
```

PR #447, opened 2022-01-21
Allow specifying the basepath in the engine settings.

marina-p:
Modified the grammar to contain a new custom payload entry
corresponding to the basepath (special keyword ```__basepath__```).

When ```basepath``` is specified in the engine settings, the engine replaces this
element with the constant specified in the settings.

If the engine settings do not specify the base path, the one from the
specification is used (via the custom payload in the generated dictionary).


marina-p:
Closes #387

PR #450, opened 2022-01-26
Support input dependencies based on annotations for any variable type.

marina-p:
Before this change, only 'restler_custom_payload_uuid4_suffix' could have
an associated input-only producer.

Now, any parameter that can be annotated may be associated with an
input-only producer.

Testing:
- Manual testing: modified demo_server to have a writer.
- Added unit test


marina-p:
This depends on PR #449

PR #503, opened 2022-04-01
Replay log fixes and add test coverage.

### LINKED ISSUE #500: replay feature is broken
This is a recent regression.  It was not caught because there are no unit tests for the replay command.

An engine unit test should be added as part of this change.

marina-p:
1. Fixed replay logic, which was broken after shared connection perf improvements.
2. Added end to end test for FuzzLean and Replay that caught the original regression.
3. After adding the two new tests, 2 other issues were found and fixed.

   a) The demo_server did not consume all the bytes of the request on all failure
   paths.  This was only caught during FuzzLean when invalid payloads were specified
   in the path and body.  The fix was to provide a teardown function that consumed
   the entire request.  The following issue has more details:
   https://github.com/pallets/flask/issues/2188

   b) test-quick-start hung in CI because the process output from demo_server was not consumed.
   This was not caught previously because there is more output in Fuzz modes.

Closes #500

### DIFF
--- a/restler/engine/core/driver.py
+++ b/restler/engine/core/driver.py
@@ -758,6 +758,14 @@ def generate_sequences(fuzzing_requests, checkers, fuzzing_jobs=1):
 
     return num_total_sequences
 
+def get_host_and_port(hostname):
+    if ':' in hostname:
+        # If hostname includes port, split it out
+        host_split = hostname.split(':')
+        return host_split[0], host_split[1]
+    else:
+        return hostname, None
+
 def replay_sequence_from_log(replay_log_filename, token_refresh_cmd):
     """ Replays a sequence of requests from a properly formed log file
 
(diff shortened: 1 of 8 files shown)


restler/engine/core/driver.py
```diff
@@ -790,6 +798,13 @@ def replay_sequence_from_log(replay_log_filename, token_refresh_cmd):
                     hostname = get_hostname_from_line(line)
                     if hostname is None:
                         raise Exception("Host not found in request. The replay log may be corrupted.")
+                    hostname, port = get_host_and_port(hostname)
+                    #if ':' in hostname:
+                    #    # If hostname includes port, split it out
+                    #    host_split = hostname.split(':')
+                    #    hostname = host_split[0]
```
marina-p, line 805: remove

PR #531, opened 2022-05-05
Update TutorialDemoServer.md

seabakmsft:
Fixed tutorial directories to match up with each other

### DIFF
--- a/docs/user-guide/TutorialDemoServer.md
+++ b/docs/user-guide/TutorialDemoServer.md
@@ -10,7 +10,7 @@ Let's create a fresh directory `C:\restler-test` where will run RESTler on this 
 
 Let's copy the Swagger spec `swagger.json` from `restler\demo_server` into `C:\restler-test` and let's compile it with RESTler:
 
-`C:\RESTler\restler\Restler.exe compile --api_spec C:\restler-test\swagger.json`
+`C:\restler-test\restler\Restler.exe compile --api_spec C:\restler-test\swagger.json`
 
 This command creates a new sub-directory `Compile` where the results of the compilation are saved in several files:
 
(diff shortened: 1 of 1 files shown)


docs/user-guide/TutorialDemoServer.md
```diff
@@ -10,7 +10,7 @@ Let's create a fresh directory `C:\restler-test` where will run RESTler on this
 
 Let's copy the Swagger spec `swagger.json` from `restler\demo_server` into `C:\restler-test` and let's compile it with RESTler:
 
-`C:\RESTler\restler\Restler.exe compile --api_spec C:\restler-test\swagger.json`
+`C:\restler-test\restler\Restler.exe compile --api_spec C:\restler-test\swagger.json`
```
marina-p, line 13: >C:\restler-test\ [](http://example.com/codeflow?start=1&length=16)

The intent here is to have separate directories for the RESTler drop directory (target directory for the built binaries) and the directory where you create the RESTler job, so these should stay different.  For clarification, what do you think about putting a sentence right before the 'Compile' section saying "The below sections assume you have built RESTler, and the RESTler drop is located in the 'C:\restler' directory.

marina-p:
Please expand the description to say why you are making these changes (just a brief explanation on having gone through the steps in the tutorial and that it was unclear where the bin directory was.)

marina-p:
> The directory `restler\demo_server` in this repo contains a self-contained example of service you can run locally on your machine. See the [README.md](/demo_server/README.md) file in that directory to install and start this demo_server service.

Related to your other feedback offline: the video (for timing reasons - it had to be under 1h) is not exactly going through the steps in the documentation.  Here, maybe we could expand this sentence to say "in that directory to set up the ```demo_server``` service.  Make sure you have started demo_server prior to going through the below steps."

---
Refers to: docs/user-guide/TutorialDemoServer.md:5 in 0122721. [](commit_id = 01227215bbe5c72ad4985e14cdd9cd132d6f690b, deletion_comment = False)

marina-p (changes requested):
:clock1:

docs/user-guide/TutorialDemoServer.md
seabakmsft: In that case, how do you feel about changing the directory to restler_bin? Under Setting up RESTler in the main readme, this is where the bin files are initially created, so it's not clear what directory "restler" refers to. Yes, I think a note under Compile will be useful as clarification in case that step was skipped as well.

marina-p: Yep, sounds good to modify c:\restler to C:\restler_bin (then, please find and replace this directory in all of the other files under 'docs', there are many other usages).

Also, since this is now a bigger change, could you please also change 'c:\restler-test' to be 'c:\demo-server-test' in this tutorial?

PR #607, opened 2022-08-17
Run the invalidvalue checker by default in fuzz modes.

marina-p:
Added baseline tests.
Added more boundary values to the invalid value generator.


restler/unit_tests/log_baseline_test_files/invalidvalue_advanced_bug_buckets.txt
```diff
@@ -0,0 +1,7 @@
+InvalidValueChecker_500: 1
+Total Buckets: 1
+-------------
+InvalidValueChecker_500 - Bug was reproduced - InvalidValueChecker_500_1.txt
+Hash: InvalidValueChecker_500_04253dc98135690c234d24a23d5a149b74d5bca6
+PUT /city/cityName HTTP/1.1\r\nAccept: application/json\r\nHost: unittest\r\nContent-Type: application/json\r\nauthentication_token_tag\r\n{"properties":{"population":10000, "area": "5000",strtest:true,"subproperties":{"subtest":true}}}\r\n
```
marina-p, line 6: >{"population":10000, "area": "5000",strtest:true,"subproperties":{"subtest":true}}}\r\n [](http://example.com/codeflow?start=153&length=87)

Double check if bucketization is working as expected.  The failing payload contains a key that is being fuzzed (in the block of strtest above).

marina-p: This will be addressed in a separate issue, since the bug is reported correctly in the repro file.

PR #608, opened 2022-08-17
wip: Add support for string ipv4 format

DeftaSebastian:
Hi, this is draft PR. I am trying to use restler on an openAPI that makes extensive use of the ipv4 string format for an internship at Keysight. I based this on your date and date-time #365 support update PR. Could you let me know if there are areas of code that might still need changing? I am also not exactly certain how the unit-tests should be run, could you also help me with that?

### DIFF
--- a/restler/engine/primitives.py
+++ b/restler/engine/primitives.py
@@ -3,6 +3,7 @@
 
 """  Primitive types supported by restler. """
 from __future__ import print_function
+import random
 import sys
 import os
 import time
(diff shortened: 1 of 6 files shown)


marina-p:
Hello @DeftaSebastian,

Thank you for contributing this!  I will get back to you with detailed feedback / instructions on adding test coverage by the end of this week.

One quick note: I've been thinking about how to improve the current design to avoid having to add a different restler_fuzzable type for every single format (e.g. having a format keyword in primitives, and a ```restler_custom_payload_format``` in the dictionary).   It would be great to implement this, and I believe it may not be much more work than adding a new type (I need to confirm this.)

Thanks,

Marina

Issue #282, opened 2021-08-10
Excluding operations


dansibleadm:
Hello everyone!

Is possible to exclude some operationId? 

Use case: I need to exclude "delete"-operation from fuzzing process. How I can do that?

Thank you!

dansibleadm:
> If you don't like commenting out, you could remove the DELETE operation from the OpenAPI specification in a pre-processing step before you compile the grammar.

That not a good way too. For some purposes I need exclude operations and then bring them back again. For example, Yelp/fuzz-lightyear support this feature and I hope RESTler support it too. As I understand, RESTler does not support it.. =(

In my opinion, the good way is exclude any modifications of source specification.

PatGod:
You can also try two other options for this:
- use the path_regex option to filter the API paths you are interested in testing -- see https://github.com/microsoft/restler-fuzzer/blob/main/docs/user-guide/SettingsFile.md#path_regex-str-default-noneno-regex-filtering and issue #269 
- edit the file grammar.py (manually or with a script) to remove all the API requests you don't want to test/fuzz.

marina-p:
Hello @dansibleadm,

RESTler does not support excluding just one of the methods today.  Can you please provide more information about your use case?  If you just exclude a DELETE for one method, do you still want to run/fuzz the corresponding PUT/POST?  If yes, is the fact that many objects will be created but not ever deleted OK for your use case?  Or is it more about creating something outside the API, then fuzzing updating it via GET/PATCH/PUT?

Thanks,
Marina

dansibleadm:
> Hello @dansibleadm,
> 
> RESTler does not support excluding just one of the methods today. Can you please provide more information about your use case? If you just exclude a DELETE for one method, do you still want to run/fuzz the corresponding PUT/POST? If yes, is the fact that many objects will be created but not ever deleted OK for your use case? Or is it more about creating something outside the API, then fuzzing updating it via GET/PATCH/PUT?
> 
> Thanks,
> Marina

My case is, for example:
I have some server with data. I need to fuzz it via OAI specification. 
1) When I create custom payloads - some resources will be deleted and part of request will not be executed. 
2) With custom payloads I need some static data about target (e.g. blogId, authorId, instanceId and etc.). When DELETE-operation was executed - I need to change my dict with new data. This is inconvenient =( Then I need to exclude delete-operation from my fuzz-process for some time.

dansibleadm:
> You can also try two other options for this:
> 
[...]

PR #614, opened 2022-08-23
Add new settings to include and exclude specific requests.

marina-p:
This is required for maintaining specific endpoints that should not be exercised, and is a cleaner way to do so than the existing path_regex switch.

Also fixes a bug in path_regex.

Testing:
- manual testing on demo_server.
- updated unit tests

Partially implements #282.  A new property will be added separately to completely exclude DELETEs from running.

### DIFF
--- a/docs/user-guide/SettingsFile.md
+++ b/docs/user-guide/SettingsFile.md
@@ -178,6 +178,8 @@ __max_examples__
 For request types where one or more examples are provided, this option limits
 the number of examples that will be tested.
 
+### max_sequence_length: int (default 100)
+The maximum length of fuzzed request sequences.
 
 ### add_fuzzable_dates: bool (default False)
 Set to True to generate additional dates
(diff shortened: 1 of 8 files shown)


docs/user-guide/SettingsFile.md
```diff
@@ -199,6 +201,39 @@ Example: `(\w*)/virtualNetworks/(\w*)`
...
+  ]
+```
+
+### exclude_requests: list (default empty list=No filtering)
+Filters the grammar to exclude the specified endpoints and methods.
+
+Note: although the ```DELETE``` is excluded from fuzzing below, it will still be executed by the RESTler
+garbage collector to clean up the blog posts that were created in order to test the other requests with
+endpoint ```/api/blog/posts/{postId}```.  To completely exclude ```DELETE```s from running, you must filter them
+manually from grammar.py.
```
marina-p, line 226: This will be improved with an option as well.

Issue #655, opened 2022-10-13
Handle 202 from DELETE requests in the garbage collector


marina-p:
### Description

The garbage collector currently accepts 202 as a deletion success code - it needs to handle 202s in the same way as in the main algorithm and retrieve the status from the ```Location:``` header.

Issue #653, opened 2022-10-13
[Feature Request] Reslter support polling delete resource operation status until it finish


ruowan:
### 💡 Idea

For azure resource, the deletion is long running operation. And it has fixed pattern to pull the deletion operation status. To avoid create too many resources at the same time, the next `create-delete` test should wait until the previous delete finish. 
restler should polling status for delete operation until the operation finish.

marina-p:
This has been implemented in #660 and #618

PR #660, opened 2022-10-20
Add async logic to DELETE

marina-p:
- Adapt existing async polling
- Add GC summary

Testing:
- manual testing with demo_server as-is and
temporarily modified to return location header on delete
- added unit test


marina-p:
Closes #655 #653

PR #662, opened 2022-10-25
Update TutorialDemoServer.md

huornlmj:
Make it clear that the demo API server needs to be already running before proceeding.


marina-p:
Thanks for the improvement!

PR #667, opened 2022-11-29
Eliminate explicit backslash path separators to fix tests on MacOS

mikekistler:
This PR fixes the compiler tests to run on MacOS by splitting up paths with explicit "\" path separators into multiple segments that are combined with Path.Combine.

In almost all cases the code was already using Path.Combine so this was relatively straightforward.

One special case was flipping a "\" to a "/" in the "examples.json" file.  I'm hoping this won't cause problems for the tests on Windows ...


marina-p:
Thank you for fixing these!

marina-p (approved):
:shipit:

PR #676, opened 2022-12-21
Remove missing file from demo server Dockerfile

tybulewicz:
logging.conf is no longer present, preventing demo image container from being built.


Yanqiulei:

PR #681, opened 2023-01-07
Add support for dependencies from OpenAPI v3 links

mikekistler:
This PR adds support for extracting dependency information from OpenAPI v3 links. Links have a similar structure to RESTler annotations. Tests and docs are included.


src/compiler/Restler.Compiler/Annotations.fs
```diff
@@ -220,4 +221,70 @@ let getGlobalAnnotationsFromFile filePath =
             printfn "ERROR: invalid annotation file: x-restler-global-annotations must be the key"
             raise (ArgumentException("invalid annotation file"))
     else
-        List.empty
\ No newline at end of file
+        List.empty
+
+// Gets the REST-ler dependency annotation from OpenAPI v3 links
```
marina-p, line 226: >REST-ler [](http://example.com/codeflow?start=12&length=8)

RESTler (without dash).  Very early versions did use REST-ler, feel free to delete dashes if you ever see them in the code base

src/compiler/Restler.Compiler/Dependencies.fs
```diff
@@ -1247,10 +1265,7 @@ let extractDependencies (requestData:(RequestId*RequestData)[])
     logTimingInfo "Getting producers..."
 
     requestData
-    // Only include POST, PUT, PATCH, and GET requests.  Others are never producers.
-    |> Array.filter (fun (r, _) -> [ OperationMethod.Post ; OperationMethod.Put; OperationMethod.Patch ;
-                                     OperationMethod.Get ]
-                                    |> List.contains r.method)
```
marina-p, line 1253: Could we please merge this and the corresponding tests separately, since this is a separate bug fix that could have wide implications on dependencies?  I'm curious why this change is needed, since the soft delete scenario has been working for me with global annotations without this change (using the annotation below).
```
    {
      "producer_endpoint": "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{vaultName}",
      "producer_method": "DELETE",
      "producer_resource_name": "vaultName",

      "consumer_endpoint": "/subscriptions/{subscriptionId}/providers/Microsoft.KeyVault/locations/{location}/deletedVaults/{vaultName}",
      "consumer_method": "GET",
      "consumer_param": "vaultName"
    
```

marina-p: Update: I haven't tested yet but I believe if you add a PUT of the config store into the soft_delete.json specs, this change should not be needed.

src/compiler/Restler.Compiler/Annotations.fs
```diff
@@ -220,4 +221,70 @@ let getGlobalAnnotationsFromFile filePath =
...
+            let consumerEndpoint = consumerOperation.Path
+            let consumerMethod = getOperationMethodFromString (consumerOperation.Method.ToString())
+            let consumerRequestId = Some {
+                                        endpoint = consumerEndpoint
+                                        method = consumerMethod
+                                        xMsPath = None
+                                    }
+            // For now we only support a single parameter
+            //let consumerParameter = link.Parameters |> Seq.tryHead |> Option.map (fun (k,v) -> ResourceName k)
+            let foo = link.Parameters |> Seq.tryHead
```
marina-p, line 266: remove

src/compiler/Restler.Compiler/Annotations.fs
```diff
@@ -220,4 +221,70 @@ let getGlobalAnnotationsFromFile filePath =
...
+            match link.Parameters |> Seq.tryHead with
+            | None ->
+                printfn "ERROR: malformed annotation, no parameters specified for link %s" link.OperationId
+                None
+            | Some kv ->
+                // Note: OAS v3 says
+                // "The parameter name can be qualified using the parameter location [{in}.]{name} for operations
+                // that use the same parameter name in different locations (e.g. path.id)."
+                let consumerParameter = Regex.Replace(kv.Key, "^(path\.|query\.|header\.|cookie\.)", "") |> ResourceName |> Some
```
marina-p, line 275: >consumerParameter [](http://example.com/codeflow?start=20&length=17)

can the consumer parameter not be in the body, or is this just not supported right now?

docs/user-guide/Links.md
```diff
@@ -0,0 +1,36 @@
+# Using OpenAPI v3 links to define dependencies
+
+The OpenAPI v3 specification defines [link objects][] that can be used to define dependencies between operations.
+Restler will use links when present in the API definition to generate dependencies between operations.
```
marina-p, line 4: >Restler will use links [](http://example.com/codeflow?start=0&length=22)

If possible, it would be great to put a short paragraph at the bottom on the supported (or not supported) subset of  Links that can be expected to work with RESTler

marina-p: In particular that only strings are supported, and operationRef and runtime expressinos are not supported

marina-p (changes requested):
:clock1:

docs/user-guide/Links.md
mikekistler: Agreed, but I have added this to Links.md rather than here.

src/compiler/Restler.Compiler/Annotations.fs
mikekistler: Wow ... what was I doing here?

src/compiler/Restler.Compiler/Annotations.fs
mikekistler: It can't -- there is no syntax for this in OpenAPI links. The consumer must be a "parameter". I've added that to the doc in Links.md.

src/compiler/Restler.Compiler/Dependencies.fs
mikekistler: I tried reverting this change and it caused my test to fail. I think this is because links are specified in the _producer_ operation rather than in the consumer, as they are for RESTler annotations. I think I have a good fix -- I just broke out the handling of links into a separate code block.

mikekistler:
@marina-p Thank you for your review/comments! I think I have addressed all of your concerns. Please review my fixes and let me know if there is anything else you'd like me to change.

src/compiler/Restler.Compiler/Dependencies.fs
marina-p: I am going to debug this a bit more to understand the root cause of the failure with the added PUT (currently not getting the expected behavior as well).  It would be good to avoid grammars that have a DELETE as a producer (without a corresponding PUT or POST first) for the time being, since the engine makes some assumptions on the producer methods as well.

mikekistler: I should clarify -- I did not try adding a PUT to my test. I only verified that the test as it exists (without a PUT) failed if DELETE was excluded from the potential producers.

And I don't understand the rationale for excluding DELETE from being a producer, and I don't recall seeing any documentation to this effect. Can you say more about what assumptions the engine makes about this?

mikekistler:
@marina-p I have commented out the logic and test for delete producers and added a new test for OpenAPI 3 links that uses a body producer. This revealed some problems with the logic that I have corrected. Please have a look and let me know if any other changes are needed.

marina-p:
Thank you for the updates.  Looks great.

---
In reply to: [1485926776](https://github.com/microsoft/restler-fuzzer/pull/681#issuecomment-1485926776) [](http://example.com/codeflow?ancestors=1485926776,1485926776)

marina-p (approved):
:shipit:

Issue #685, opened 2023-01-16
NullReferenceException when running test


lorg:
### Description

While trying to run restler on the spec of a large project, I ran into a nullreference exception. Here is the command:
`dotnet "/Users/ [...snip...] /fuzzing/restler-fuzzer/restler_bin/restler/Restler.dll"
test --grammar_file "Compile/grammar.py" --dictionary_file "Compile/dict.json" --settings "Compile/engine_settings.json"`

Here is the output:
```
Starting task Test...
Using python: 'python3' (Python 3.10.8)
Request coverage (successful / total): 0 / 42
Attempted requests: 0 / 42
No bugs were found.

ERROR: Task Test failed.  Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Restler.Driver.SpecCoverage.getEscapedString@320-1.Invoke(String state, Tuple`2 tupledArg) in /Users/ [ ....snip...] /fuzzing/restler-fuzzer/src/driver/SpecCoverage.fs:line 321
   at Microsoft.FSharp.Collections.ListModule.Fold[T,TState](FSharpFunc`2 folder, TState state, FSharpList`1 list) in F:\workspace\_work\1\s\src\fsharp\FSharp.Core\list.fs:line 222
   at Restler.Driver.SpecCoverage.getRequestResponseText@316(RequestCoverageData reqData) in /Users/ [ .... snip ... ] /fuzzing/restler-fuzzer/src/driver/SpecCoverage.fs:line 326
   at Restler.Driver.SpecCoverage.getRequestSequence(FSharpList`1 reqCoverageData, FSharpMap`2 renderingToReqCovMap, RequestCoverageData renderedCombination) in /Users/ [ .... snip ... ] /fuzzing/restler-fuzzer/src/driver/SpecCoverage.fs:line 343
   at Restler.Driver.SpecCoverage.combinations@421.Invoke(RequestCoverageData renderedCombination)
   at Microsoft.FSharp.Collections.Internal.IEnumerator.map@75.DoMoveNext(b& curr) in F:\workspace\_work\1\s\src\fsharp\FSharp.Core\seq.fs:line 78
   at Microsoft.FSharp.Collections.Internal.IEnumerator.MapEnumerator`1.System-Collections-IEnumerator-MoveNext() in F:\workspace\_work\1\s\src\fsharp\FSharp.Core\seq.fs:line 64
[...]

PR #686, opened 2023-01-18
Handle null request or response properties in speccov parsing

marina-p:
- Null requests are filtered out (it means the request was not executed)
- Null responses are included, this may be due to connection failures caused by the request and should
be shown in the coverage investigation file

Testing:
- manual testing using the repro in the reported issue #685


marina-p:
Closes #685

Issue #692, opened 2023-01-27
RESTler should try to get the resource when async polling fails


marina-p:
### Description

Today, RESTler tries to get the resource (by constructing a GET corresponding to a POST) only if async polling times out.  However, it is possible that polling fails due to a bug (e.g. incorrect location).  RESTler should log these errors, but they should not block fuzzing.  Similarly to a polling timeout, RESTler should attempt to use a GET to obtain the resource ID that is supposed to be present in the POST response.

marina-p:
Closing since the fix was merged.

PR #693, opened 2023-01-27
Polling fix - #692

marina-p:
Uses GET to try to obtain dynamic object IDs for any polling failure.

### DIFF
--- a/restler/engine/core/async_request_utilities.py
+++ b/restler/engine/core/async_request_utilities.py
@@ -178,11 +178,11 @@ def try_async_poll(request_data, response, max_async_wait_time, poll_delete_stat
                     # Send the polling request
                     poll_response = request_utilities.send_request_data(data)
                     time_str = str(round((time.time() - start_time), 2))
-                    if data_in_poll_response:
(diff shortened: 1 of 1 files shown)


restler/engine/core/async_request_utilities.py
```diff
@@ -198,9 +199,16 @@ def try_async_poll(request_data, response, max_async_wait_time, poll_delete_stat
                                     responses_to_parse.insert(0, get_response)
                             # Break and return the responses to be parsed
                             break
+                        elif int(poll_response.status_code) >= 300:
+                            # Error obtaining the polling response.
+                            # It may still be possible to extract the relevant ID out of the response
+                            responses_to_parse.append(poll_response)
```
abarton425, line 205: poll_response in this case is the 404 not found. If we simply break and don't add responses_to_parse in this else if, the if at line 267 will automatically add the original response, which will bypass the 404 all together and resolve this issue.

marina-p: I see, thanks, I had mis-copied that from your logs.  I'll make the change shortly, please let me know when you've tried it to confirm so I can merge.

Issue #694, opened 2023-01-27
[Feature Request] Restler should test presence of auth files as part of settings validation


ajpasupu:
### 💡 Idea

Restler should test for presence of auth module if module authentication is specified during settings validation and fail fast if it is unreachable/does not exist

ajpasupu:
Closed with https://github.com/microsoft/restler-fuzzer/pull/698

PR #695, opened 2023-01-27
Don't print duplicate module name

ajpasupu:
## PR Checklist
* [x] Applies to work item: N/A very small bug fix 
* [x] Tests added/passed.  Tests passed. I think it would make sense to add this validation as part of restler settings validation and add a unit test for that, rather than adding e2e tests to catch this case. Created an issue for this https://github.com/microsoft/restler-fuzzer/issues/694
* [x] Requires documentation to be updated. No 
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different approach. Issue number where discussion took place: #xxx

### DIFF
--- a/restler/engine/core/request_utilities.py
+++ b/restler/engine/core/request_utilities.py
@@ -141,13 +141,12 @@ def execute_token_refresh_module(module_path, function, data):
     @return: token
     @type: string:
     """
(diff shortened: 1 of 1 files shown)


marina-p (approved):
:shipit:

PR #698, opened 2023-02-06
Add check for module file existence as setting validation

ajpasupu:
## PR Checklist
* [x] Applies to work item: https://github.com/microsoft/restler-fuzzer/issues/694
* [x] Tests added/passed.  
* [x] Requires documentation to be updated (No)
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different approach. Issue number where discussion took place: #xxx


marina-p (approved):
:shipit:

PR #737, opened 2023-04-22
correct the wrong pagination query

Lucarun:
"select * from blogpost limit 5 offset 10" means querying 5 rows of data starting from row 10 in the sqlite. It doesn't align with the actual meaning of this pagination query API - "/api/blog/posts". (page and per_page). Please correct me if I am wrong. Thx!


marina-p:
You are correct.  I'm going to check the previous implementation of demo_server before merging it to see if this was supposed to be a planted bug, or if I introduced this bug unintentionally while porting it.

Yanqiulei:

PR #738, opened 2023-04-22
correct the HTTPException msg

Lucarun:
if payload.id < 10, the HTTPException msg should be "ID must be at least 10"


marina-p:
Thanks for the fix!

marina-p (approved):
:shipit:

Issue #756, opened 2023-05-09
[Feature Request] Allow static User-Agent header value


corygehr:
### 💡 Idea

Some API endpoints will only accept requests if User-Agent criterion are met. A setting which allows users to provide a static value for the User-Agent header would make fuzzing these endpoints possible.

### Design Notes

It seems like RESTler may have certain overrides for this specific header today (for example, a setting exists to decide whether or not User-Agent should be sent at all in requests). This may need to be a separate option entirely to ensure a custom value is included in all areas where requests could be sent.

PR #761, opened 2023-05-18
Allow specifying a custom user agent.

marina-p:
A custom user agent may now be specified in the engine settings.

Testing:
- manual testing with demo_server


marina-p:
Closes #756

PR #767, opened 2023-05-23
Allow specifying Content-Length as a custom payload

marina-p:
Previously, ```Content-Length``` could only be specified as a ```restler_custom_payload_header```, which injects the header for every request.  This change allows ```Content-Length``` to also be specified as a custom payload, which only includes it in requests that declare it in the specification.


Oleggg2000:
Hi @marina-p is it possible to provide me with en example of adding Content-Length or Content-Type (that what i need) for specific request. The way showing in img isn't working for me. I might think that i've already seen that "restler_custom_payload" is **replacing** only and "restler_custom_payload_header" could add en extra header, but that parameter is adding for every request. What could u suggest me? 
Apprisiate for future replay.
image
specification
image

PR #784, opened 2023-07-05
typo on how settings singleton is called

alaendro:


marina-p (approved):
:shipit:

marina-p:
Thanks for the fix!

Issue #699, opened 2023-02-07
restler-fuzzer connect server with http rather than https


NanFeng009:
### Description

In swagger.json, i have server like this format  

> servers:
>   - url: https://xx.xx.xx.xx:xxxx/example/

After compile, in grammar.py, i saw host like this 

> primitives.restler_static_string("Host: xx.xx.xx.xx:xxxx\r\n"),

In grammar.json, i saw this 

>         [
>           "Host",
>           "xx.xx.xx.xx:xxxx"
>         ]

After test:
I got "Request coverage (successful / total): 0 / 7" in RestlerLogs/restler-20230206-100623.log
got "`**Empty response received.  Re-creating connection and re-trying.**`" in Test/RestlerResults/experiment265302/logs/network.testing.140076775822912.1.txt

Used wireshark, i saw all requests were sent out in http rather than https.

NanFeng009:
I found the demo server also have similar issue 
I setup demo_server following https://github.com/microsoft/restler-fuzzer/blob/main/docs/user-guide/TutorialDemoServer.md
and I saw the swagger.json has host is **https**
`  "servers": [
        {
            "url": "https://localhost:8888",
            "description": "Staging environment"
        }
`
but from the sniffer i saw, the fuzzer connect server with http rather than https
![image](https://user-images.githubusercontent.com/13536178/217731235-3eaad482-55f4-4bfa-a441-3c273a2f6c91.png)

marina-p:
Hello @NanFeng009,

For your own service, could you please share more details about your scenario?  Is authentication configured for your service?  There may be several reasons for the "empty response received" error.  If it is possible for you to attach the same trace for a working trace of your service (you can remove the endpoints), that would be helpful.

The `demo_server` traffic you captured looks correct to me.  The Swagger file was auto-generated by FastAPI, good catch that is a bug in the `demo_server` code and it should be `http://localhost:8888`.  However, the `https` is stripped off for the host name (as you saw in the grammar), so there is no issue connecting.

Thanks,

Marina

leonasdev:
> The `demo_server` traffic you captured looks correct to me. The Swagger file was auto-generated by FastAPI, good catch that is a bug in the `demo_server` code and it should be `http://localhost:8888`. However, the `https` is stripped off for the host name (as you saw in the grammar), so there is no issue connecting.

Hello @marina-p,

[...]

PR #806, opened 2023-08-24
Fix `demo_server` should using http instead of https

leonasdev:
close https://github.com/microsoft/restler-fuzzer/issues/699


marina-p:
Thank you for the fix!

marina-p (approved):
:shipit:

PR #813, opened 2023-09-26
Fix Dockerfile

dnoliver:
* The logging.conf file is not there by default

Hi, just found this problem in the demo server, removing the logging.conf file from the demo server Dockerfile so it builds


marina-p:
Thank you for the fix!

marina-p (approved):
:shipit:

PR #829, opened 2023-11-02
Enable structured logging for RESTler request sequences

marina-p:
This change adds structured logging for all of the network traffic that is currently logged to the network.*.txt log.  The data is written to a 'trace database'. The default format is newline-delimited json, but custom logging formats are supported via a module specified in the engine settings (similar to checkers).

Below is an example for demo_server:
```
{"sent_timestamp": "2023-11-02T10:10:59.544824+00:00", "received_timestamp": null, "request": "GET /api/blog/posts?page=1&per_page=1 HTTP/1.1\r\nAccept: application/json\r\nHost: localhost\r\nContent-Length: 0\r\nx-restler-sequence-id: fe2172ab-151c-419f-b918-f3e7483b3230\r\n\r\n", "response": null, "request_json": "null", "response_json": null, "tags": {"request_id": "27f9653431313fdc3fecc4a890b72b80b4ce1e59", "sequence_id": "fe2172ab-151c-419f-b918-f3e7483b3230", "combination_id": "1950cbddab7726489624c3d346d3426561c921ad_1", "hex_definition": "6daf8d22c7a6b3472fc83c9f08f290b3507c3ff3", "origin": "main_driver"}}

{"sent_timestamp": null, "received_timestamp": "2023-11-02T10:10:59.597613+00:00", "request": null, "response": "HTTP/1.1 400 Bad Request\r\ndate: Thu, 02 Nov 2023 10:10:58 GMT\r\nserver: uvicorn\r\ncontent-length: 41\r\ncontent-type: application/json\r\n\r\n{\"detail\":\"per_page must be at least 2.\"}", "request_json": null, "response_json": null, "tags": {"request_id": "27f9653431313fdc3fecc4a890b72b80b4ce1e59", "sequence_id": "fe2172ab-151c-419f-b918-f3e7483b3230", "combination_id": "1950cbddab7726489624c3d346d3426561c921ad_1", "hex_definition": "6daf8d22c7a6b3472fc83c9f08f290b3507c3ff3", "origin": "main_driver"}}
```

Testing:
- demo_server
- added new unit test
- manual testing for multiple db files


restler/restler_settings.py
```diff
@@ -441,6 +445,8 @@ def convert_wildcards_to_regex(str_value):
         self._ignore_feedback = SettingsArg('ignore_feedback', bool, False, user_args)
         ## Include user agent in requests sent
         self._include_user_agent = SettingsArg('include_user_agent', bool, True, user_args)
+        ## Include a unique sequence ID in requests sent
```
wilbaker, line 448: For my own understanding, if this were set to false, would it be possible to determine which requests in the structured log were for the same sequence? (It looks like requests/responses could be linked using `combination_id`? But I wanted to confirm if that is correct).

restler/utils/logging/trace_db.py
```diff
@@ -0,0 +1,225 @@
...
+            # print the callstack
+            import traceback
+            traceback.print_exc()
+            print(f"Warning: Exception logging request/response: {error}")
+            pass
+
+    def load_trace_data(self):
+        self.storage_writer.load()
+
+    def normalize_trace_data(self):
```
wilbaker, line 148: Is the `normalize_trace_data` method used anywhere today, or is this intended to be called once the TODO in ndjson serializer's `normalize()` implementation is complete?

(Asking because I couldn't find where this method is being used, and I wasn't sure if I missed it)

restler/utils/logging/trace_db.py
```diff
@@ -0,0 +1,225 @@
...
+        if tls_trace_log is None:
+            return None
+        return tls_trace_log.sequence_id
+
+    def log_request_response(self, request=None, response=None,  tags={}, timestamp=None):
+        if request is None and response is None:
+            raise Exception("ERROR: Request or response must be specified.")
+        try:
+            tls_trace_log = get_trace_log()
+            if tls_trace_log is None:
```
wilbaker, line 118: For my own understanding, how does this condition work (checking if `get_trace_log` returns None to determine if the request is outside the sequence context)? Do checkers run in their own thread (that do not initialize `sequence_trace`)?

wilbaker (commented):
Still reviewing the changes, but I had a few questions to help my understand how the changes work, thanks!

restler/restler_settings.py
marina-p: Yes, it is still possible.  This setting is independent of anything in the structured log, and applies to the payloads themselves.  I should probably split this off to its own PR which will go in after this one.  The same 'sequence_id' is also available in the structured log by default, in the tags.  'combination_id' is the grammar-defined combination, so it may appear twice if you are running in random mode, but the 'sequence_id' is a unique guid generated for every sequence run.

restler/utils/logging/trace_db.py
marina-p: Correct, it is not used yet.  Will add a TODO.

restler/utils/logging/trace_db.py
marina-p: Garbage collection runs in its own thread.  Checkers do not run in their own thread - they run after the "main" sequence has finished.  However, checkers in general can have custom logic and may decide not to run 'initialize_sequence_trace', which this logic would cover.

marina-p: This PR iteration does not yet have the checker logic fully working for tracking sequences.

marina-p:
Just fixed a few issues - please get the latest iteration for testing

---
In reply to: [1711381128](https://github.com/microsoft/restler-fuzzer/pull/829#issuecomment-1711381128) [](http://example.com/codeflow?ancestors=1711381128)

wilbaker:
> Just fixed a few issues - please get the latest iteration for testing
> 
> In reply to: [1711381128](https://github.com/microsoft/restler-fuzzer/pull/829#issuecomment-1711381128)

Thanks! I will give this a try.

restler/utils/logging/trace_db.py
```diff
@@ -0,0 +1,225 @@
...
+        obj = custom_serializer(Settings().trace_db_serializer_settings)
+    return obj
+
+def set_up_trace_database(root_dir_path=None):
+    global _db
+    if _db is not None:
+        raise Exception("Trace database already initialized")
+
+    if root_dir_path is None:
+        root_dir_path = os.getcwd()
```
wilbaker, line 169: I wanted to confirm if this was intentional, I noticed that by default the trace_db file is not created in the same directory as the per-experiment logs.

I am seeing trace_data.ndjson created in the Fuzz directory, whereas the network logs are in Fuzz/RestlerResults/experimentXYZ/logs

marina-p: It was intentional, but thinking about this again, it would be cleaner to put this under the experiment directory by default.

restler/utils/logging/trace_db.py
marina-p: The checkers are invoked after the main driver sequence has finished running (see driver.py line 210-217).  So first the main driver will initialize and clear the sequence trace (see 'clear_trace_log'), and then the checker will start a new sequence trace (since the checker will re-render each sequence to which it is being applied from scratch).  

The main driver should always clear the trace log before passing control to the checker.  I will try to rename 'clear_trace_log' to better reflect the above intent.

marina-p:
TODO: test replay mode with trace DB & update replay sequence to generate a unique sequence ID.

restler/utils/logging/trace_db.py
wilbaker: Ahh okay, thank!

restler/utils/logging/ndjson_serializer.py
```diff
@@ -0,0 +1,183 @@
...
+        self._response = value
+
+    def to_dict(self):
+        tags = {}
+        if self.request_id:
+            tags["request_id"] = self.request_id
+        if self.sequence_id:
+            tags["sequence_id"] = self.sequence_id
+        if self.combination_id:
+            tags["combination_id"] = self.combination_id
```
wilbaker, line 99: I'm not sure if this is the right place for this comment, but would it be possible to include the origin too (e.g., "main_driver" or the checker's name)?

Apologies for not posting this comment sooner, I thought I had added it last week but I couldn't find it this morning.

marina-p: Yep!  Already planning to do this for the next iteration.  Let me know if you think of any others that would be useful.

restler/utils/logging/ndjson_serializer.py
```diff
@@ -0,0 +1,183 @@
...
+        tags.update(self.tags)
+
+        return {
+            'sent_timestamp': self.sent_timestamp,
+            'received_timestamp': self.received_timestamp,
+            'request': self.request,
+            'response': self.response,
+            'request_json': json.dumps(self.request_json),
+            'response_json': json.dumps(self.response_json),
```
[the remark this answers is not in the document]
marina-p: Thanks for catching, this is a bug and should be `null` without the quotes.

restler/utils/logging/ndjson_serializer.py
wilbaker: > Let me know if you think of any others that would be useful.

@marina-p, related to this request, while testing I found that the garbage collection thread is logging its requests and responses into the trace database. I tracked this down while investigating why my trace database had so many DELETE request/response pairs logged without `sequence_id`. I have `run_gc_after_every_sequence` set to true, which makes this easy to repro.

Assuming that the plan is to keep the GC thread activity included in the trace database, it would be helpful if there were an easy way to identify that these lines are coming from the GC thread. Is this something that could be included with the change to log the origin (maybe have a specific origin for the garbage collector)?

Thanks!

marina-p: Yep, will include this with the origin change.

restler/utils/logging/trace_db.py
marina-p: Moved to ExperimentXYZ

restler/engine/core/sequences.py
```diff
@@ -777,6 +806,12 @@ def send_and_parse(request_data):
                 request_utilities.call_response_parser(request_data.parser, None, responses=responses_to_parse)
             return response.status_code
 
+        # TODO: when replaying from the trace DB is supported, the combination ID should be available, since
+        # the sequence object will be available.  It may also be useful to provide the original combination ID
+        # for the sequence, so that the replay can be associated with the original rendering.
```
wilbaker, line 811: Not for this PR, but once this feature is supported it would be helpful if the original origin (e.g., "main_driver") could be recorded as a tag as well.

wilbaker (approved):

PR #839, opened 2023-11-16
Enable generating unique sequence IDs.

marina-p:
Add a setting that, when true, sends a header `x-restler-sequence-id` with each request containing a unique GUID that identifies the RESTler sequence.  

This setting is enabled by default.

Testing:
- manual testing

### DIFF
--- a/restler/engine/transport_layer/messaging.py
+++ b/restler/engine/transport_layer/messaging.py
@@ -8,7 +8,8 @@ import socket
 import time
 import threading
 from importlib import util
-from utils.logging.trace_db import DB as TraceDatabase
+from utils.logging.trace_db import (DB as TraceDatabase,
+                                    SequenceTracker)
 from utils.formatting import iso_timestamp
 from utils.logger import raw_network_logging as RAW_LOGGING
 from engine.errors import TransportLayerException
@@ -200,6 +201,10 @@ class HttpSock(object):
         elif self.connection_settings.include_user_agent:
             # Send the RESTler user agent only if a custom user agent is not specified
             message = _append_to_header(message, f"User-Agent: restler/{Settings().version}")
+        if self.connection_settings.include_unique_sequence_id:
+            sequence_id = SequenceTracker().get_sequence_id()
+            if sequence_id is not None:
+                message = _append_to_header(message, f"x-restler-sequence-id: {sequence_id}")
 
         # Attempt to throttle the request if necessary
         self._begin_throttle_request()
--- a/restler/utils/logging/trace_db.py
+++ b/restler/utils/logging/trace_db.py
@@ -5,49 +5,46 @@
 
 from restler_settings import Settings
 from utils.logging.ndjson_serializer import *
-import os
 import time
-import json
 import queue
 import threading
 import uuid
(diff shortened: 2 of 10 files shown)


wilbaker:
@marina-p changes look good to me, I am running a manual test of the changes now

wilbaker (approved):
Changes look good, but I noticed that the unit tests are currently failing.

restler/engine/transport_layer/messaging.py
```diff
@@ -200,6 +201,8 @@ def _append_to_header(message, content):
         elif self.connection_settings.include_user_agent:
             # Send the RESTler user agent only if a custom user agent is not specified
             message = _append_to_header(message, f"User-Agent: restler/{Settings().version}")
+        if self.connection_settings.include_unique_sequence_id:
+            message = _append_to_header(message, f"x-restler-sequence-id: {SequenceTracker().get_sequence_id()}")
```
wilbaker, line 205: @marina-p actually, I do have a comment here after taking a closer look at the logging from my test run.

Requests sent from the garbage collector will include `x-restler-sequence-id: None` in the header. Is that desirable, or should `x-restler-sequence-id` only be added when there is a non-None sequence ID?

marina-p: It should be sent only if present - thanks for catching.

marina-p:
Thanks, fixing the test in the next update.  This was an interesting and good catch - updated RESTler code to use a separate Random() object for sequence generation so it doesn't interfere with the custom user-defined random generator 

---
In reply to: [1735409793](https://github.com/microsoft/restler-fuzzer/pull/839#issuecomment-1735409793) [](http://example.com/codeflow?ancestors=1735409793)

restler/utils/logging/trace_db.py
```diff
@@ -12,55 +12,58 @@
...
 import uuid
 import utils.import_utilities as import_utilities
-import random
+from random import Random
 
 DEFAULT_ORIGIN = 'main_driver'
 
 # Thread local storage for tracking the current sequence and request
 threadLocal = threading.local()
+trace_db_random_gen = Random(Settings().random_seed)
```
wilbaker, line 21: Looks like tests are failing here:

```
engine/dependencies.py:15: in <module>
    from utils.logging.trace_db import SequenceTracker
utils/logging/trace_db.py:21: in <module>
    trace_db_random_gen = Random(Settings().random_seed)
restler_settings.py:343: in Settings
    return RestlerSettings.Instance()
restler_settings.py:357: in Instance
    raise UninitializedError("RestlerSettings not yet initialized.")
E   restler_settings.UninitializedError: RestlerSettings not yet initialized.
```

marina-p: Removing this code so it's not blocking this change.

marina-p: Just FYI, I do not run all the tests locally before pushing changes.  All the failures are required to be fixed before submission.

wilbaker: Ok thanks! To reduce noise going forward I won't leave comments related to validation pipeline failures.

wilbaker (approved):

PR #847, opened 2024-01-17
Compiler: fix ref unescaping

justdanpo:
Reference to path works incorrectly when path contains more than one slash (see example below), compiler cannot find such path:
```log
Unhandled exception. System.InvalidOperationException: Reference path "/paths/~1xxx~1ab" not found
```

Also current unescaping breaks reference part into an extra level i.e. reference `#/definitions~1Escape~0Characters` points to
`Escape~Characters` inside `definitions` while it should point to `definitions/Escape~Characters` in the root.

This PR fixes the issue by performing unescaping after splitting path to parts.

Example schema:
```yaml
openapi: 3.0.3
info:
  title: REST API
  description: REST API description
  version: 1.0.0
servers:
  - url: http://{hostname}:{port}
    variables:
      hostname:
        default: localhost
      port:
        default: "80"

paths:
  /api/device/info:
    $ref: "#/paths/~1xxx~1ab"

  /xxx/ab:
    get:
      summary: Get information
      description: Return identifier
      responses:
        '200':
          description: info
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Info'

components:
  schemas:
    Info:
      type: object
      required:
        - id
      properties:
        id:
          type: string
```

### DIFF
--- a/src/compiler/Restler.Compiler.Test/swagger/schemaTests/swagger_escape_characters.json
+++ b/src/compiler/Restler.Compiler.Test/swagger/schemaTests/swagger_escape_characters.json
@@ -14,7 +14,7 @@
             "required": true,
             "in": "body",
             "schema": {
-              "$ref": "#/definitions~1Escape~0Characters"
+              "$ref": "#/definitions/Escape~0Characters"
             }
           }
         ],
@@ -22,7 +22,7 @@
           "200": {
             "description": "Success",
             "schema": {
-              "$ref": "#/definitions~1Escape~0Characters"
+              "$ref": "#/definitions/Escape~0Characters"
             }
           }
         }
@@ -45,7 +45,7 @@
           "200": {
             "description": "Success",
             "schema": {
-              "$ref": "#/definitions/~01EscapeCharacters"
+              "$ref": "#/definitions/~1Escape~1Characters"
             }
           }
         }
@@ -96,6 +96,13 @@
     "~1EscapeCharacters": {
       "properties": {
         "escape_characters2": {
+          "type": "string"
+        }
+      }
+    },
+    "/Escape/Characters": {
+      "properties": {
+        "escape_characters3": {
           "type": "string"
         }
       }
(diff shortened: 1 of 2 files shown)


justdanpo:
@microsoft-github-policy-service agree

Yanqiulei:

src/compiler/Restler.Compiler.Test/swagger/schemaTests/swagger_escape_characters.json
```diff
@@ -14,15 +14,15 @@
             "required": true,
             "in": "body",
             "schema": {
-              "$ref": "#/definitions~1Escape~0Characters"
+              "$ref": "#/definitions/Escape~0Characters"
```
marina-p, line 17: >ions/Escape~0Cha [](http://example.com/codeflow?start=32&length=16)

@justdanpo Thank you for the fix!  Apologies for the delay in merging this.  Just one quick question, why did you modify these existing refs - are they not valid?  If they are still expected to work, could you please keep them and add separate operations for the new ones?

marina-p (changes requested):
:clock1:

src/compiler/Restler.Compiler.Test/swagger/schemaTests/swagger_escape_characters.json
justdanpo: `"$ref": "#/definitions~1Escape~0Characters"` points to
```json
{
  "definitions/Escape~Characters": {
    "properties": {
      "escape_characters1": {
        "type": "string"
      }
   }
}
```
which does not exist in the schema.
Correct reference `"$ref": "#/definitions/Escape~0Characters"` points to
```json
{
  "definitions": {
    "Escape~Characters": {
      "properties": {
        "escape_characters1": {
          "type": "string"
        }
      }
    }
}
```

marina-p (approved):
:shipit:

PR #852, opened 2024-02-01
Change JsonSerializer to require all non-optional fields

eddyashton:
I was surprised by how permissive the JSON deserialization is, and it resulted in error messages hiding useful information.

My motivating use case was an annotations file like this:
```
{
  "x-restler-global-annotations": [
    { }
  ]
}
```

Trying to compile a grammar with that produces an extremely generic null-reference message:
> Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Restler.Annotations.getAnnotationsFromJson(JToken annotationJson) in /data/src/restler-fuzzer/src/compiler/Restler.Compiler/Annotations.fs:line 168
   at Restler.Annotations.getGlobalAnnotationsFromFile(String filePath) in /data/src/restler-fuzzer/src/compiler/Restler.Compiler/Annotations.fs:line 220
   at Restler.Workflow.generateGrammarFromSwagger(String grammarOutputDirectoryPath, Config config) in /data/src/restler-fuzzer/src/compiler/Restler.Compiler/Workflow.fs:line 104
   at Restler.Workflow.generateRestlerGrammar(Config config) in /data/src/restler-fuzzer/src/compiler/Restler.Compiler/Workflow.fs:line 250
   at Program.main(String[] argv) in /data/src/restler-fuzzer/src/compiler/Restler.CompilerExe/Program.fs:line 43

It turns out that even though `ProducerConsumerUserAnnotation` says `producer_endpoint : string` (implying a non-optional field), it will parse `{}` and produce an object with `{producer_endpoint: null}`. Despite sensible checks on the result of `tryDeserialize`, this is missed and blows up later. I think this should be a parsing error, caught by all callers of `Utilities.JsonSerialization.tryDeserialize` and reported correctly.

This PR makes that change, by specifying a `ContractResolver` from `FSharpLu.Json`. The original example now produces a descriptive error:

> Unhandled exception. Newtonsoft.Json.JsonSerializationException: Required property 'producer_endpoint' not found in JSON. Path '', line 1, position 2.
...

(Disclaimer - I'm not familiar with F# or .NET, and I've only just started experimenting with restler, so this is a purely speculative change from a bit of docs-probing. Please let me know if there's a better way to achieve this, or some reason that the original behaviour is required)

### DIFF
--- a/src/compiler/Restler.Compiler/Utilities.fs
+++ b/src/compiler/Restler.Compiler/Utilities.fs
@@ -57,7 +57,8 @@ module JsonSerialization =
                 JsonSerializerSettings(
                     NullValueHandling = NullValueHandling.Ignore,
                     MissingMemberHandling = MissingMemberHandling.Error,
-                    DateParseHandling = DateParseHandling.None
+                    DateParseHandling = DateParseHandling.None,
+                    ContractResolver = Compact.Strict.RequireNonOptionalPropertiesContractResolver()
                 )
             settings.Converters.Add(CompactUnionJsonConverter(true, true))
             settings


eddyashton:
CI failed, but it looks like this was due to a race condition, concurrently accessing files:

```
  Failed Restler.Test.Dictionary+DictionaryTests.quoting is correct in the grammar [1 s]
  Error Message:
   System.IO.IOException : The process cannot access the file 'C:\Users\VssAdministrator\AppData\Local\Temp\restlerTest\202142\grammar.json' because it is being used by another process.
  Stack Trace:
    ...
  Failed Restler.Test.Dictionary+DictionaryTests.path and body parameter set to the same uuid suffix payload [30 ms]
  Error Message:
   System.IO.IOException : The process cannot access the file 'C:\Users\VssAdministrator\AppData\Local\Temp\restlerTest\202142\grammar.json' because it is being used by another process.
```

Is this a known flaky test? Should the CI simply be re-run?

marina-p:
> CI failed, but it looks like this was due to a race condition, concurrently accessing files:
> 
> ```
>   Failed Restler.Test.Dictionary+DictionaryTests.quoting is correct in the grammar [1 s]
>   Error Message:
>    System.IO.IOException : The process cannot access the file 'C:\Users\VssAdministrator\AppData\Local\Temp\restlerTest\202142\grammar.json' because it is being used by another process.
>   Stack Trace:
>     ...
>   Failed Restler.Test.Dictionary+DictionaryTests.path and body parameter set to the same uuid suffix payload [30 ms]
>   Error Message:
>    System.IO.IOException : The process cannot access the file 'C:\Users\VssAdministrator\AppData\Local\Temp\restlerTest\202142\grammar.json' because it is being used by another process.
> ```
> 
> Is this a known flaky test? Should the CI simply be re-run?

Correct, this failure is a known race condition and it's fine to merge if the test passes on re-run.

src/compiler/Restler.Compiler/Utilities.fs
```diff
@@ -57,7 +57,8 @@ module JsonSerialization =
                 JsonSerializerSettings(
                     NullValueHandling = NullValueHandling.Ignore,
                     MissingMemberHandling = MissingMemberHandling.Error,
-                    DateParseHandling = DateParseHandling.None
+                    DateParseHandling = DateParseHandling.None,
+                    ContractResolver = Compact.Strict.RequireNonOptionalPropertiesContractResolver()
```
marina-p, line 61: >RequireNonOptionalPropertiesContractResolver [](http://example.com/codeflow?start=54&length=44)

This type from the FSharpLU project inherits from CamelCasePropertyContractResolver, which is not desired here.  You could define your own type in this 'JsonSerialization' module that is identical except inherits from the default contract resolver:

```
        /// A contract resolver that requires presence of all properties
        /// that are not of type option<_>
        type RequireNonOptionalPropertiesContractResolver() =
            inherit Newtonsoft.Json.Serialization.DefaultContractResolver()
            override __.CreateProperty(_member, memberSerialization) =
                let property = base.CreateProperty(_member, memberSerialization)
                let isRequired = not (property.PropertyType.GetTypeInfo().IsGenericType 
                                   && property.PropertyType.GetGenericTypeDefinition() = typedefof<option<_>>)
                if isRequired then 
                    property.Required <- Required.Always
                    property.NullValueHandling <- System.Nullable NullValueHandling.Ignore
                property

```

marina-p (changes requested):
:clock1:

src/compiler/Restler.Compiler/Utilities.fs
eddyashton: It looks like the base type [now inherits from `DefaultContractResolver`](https://github.com/microsoft/fsharplu/blob/805dce44eb8b4110a1a03ce835a3fe74dcc0eaff/FSharpLu.Json/Compact.fs#L439) (and has since [2019](https://github.com/microsoft/fsharplu/commit/5ff9741f073d81ba51ed445bfa66d009b457142a), where camelCasing was moved from `NonOptional...` to `CompactStringSettings`), so this local definition exactly matches the one upstream.

To confirm, I tried a local definition which inherits from `CamelCasePropertyContractResolver` - that fails several unit tests. So the version in this PR, which passes those unit tests, must be using `DefaultContractResolver` already.

Happy to move this definition inline if there's a reason not to rely on the upstream, but it looks unnecessary (and protected by unit tests) at the moment.

marina-p: Ah, my bad!  I must have looked at a very old version on GitHub, not sure how that happened.  Yes, looks good, no reason to move the definition inline.

marina-p (approved):
:shipit:

Issue #632, opened 2022-09-20
ResultsAnalyzer could not parse request


taylorThunderbolt:
### Description

Hi, 
I start restler in the fuzz-lean mode. The command is as follows:
```bash
..\restler_bin\restler\Restler.exe fuzz-lean --grammar_file Compile\grammar.py --dictionary_file Compile\dict.json --settings Compile\engine_settings.json --token_refresh_interval 30 --token_refresh_command "python3 get_token.py" --no_ssl
```
After a period of time, the ResponseBuckets directory is created, but it is empty.There are a bunch of lines in the ResultsAnalyzerStdErr file like:
```bash
Warning: Warning: could not parse request in line 9585: GET /api/alarms/rules?pageNo=1&pageSize=1&sortBy=host&orderBy=host&filter=host&search==B?G|N=g2DVP@n+VmDbJ0aVq}4+kF?v&&\r_^2sqh)r"s5"wR^69Tl HTTP/1.1\r\nAccept: application/json\r\nHost: 192.168.150.137:9060\r\n_OMITTED_AUTH_TOKEN_\r\nContent-Length: 0\r\nUser-Agent: restler/9.0.1\r\n\r\n
Unhandled exception. System.Exception: Unexpected response without prior request at 2022/9/20 15:29:17: { version = "HTTP/1.1"
  statusCode = 400
  statusDescription = "Bad Request"
  headers =
           map
             [("Connection", "close");
              ("Content-Type", "text/plain; charset=utf-8")]
  body = "400 Bad Request"
  str = None }
```
What does it mean?

marina-p:
This is a known issue that should be fixed soon (also reported in #465).

taylorThunderbolt:
Your reply helps me a lot, thank u ^_^

tandrew-test:
I am seeing this issue too. I've had to write my own parser to ensure that I do not miss anything.

Here are some sample urls that appear to be causing issues when in the log:
* `v2/widget/8YIo@RL7lu:x2x"jj%^^f`
* `v2/widget/k7044"U+_}avUm-~`0O.[>x{wKu?*|4v8A>a":}/`

I do not have any trouble parsing the logs using Python so it's something to do with the F# implementation.

PR #864, opened 2024-02-22
updating regex to match logs expression

Alebmurillo:

 This PR aims to solve the Resutls analyzer issue [632](https://github.com/microsoft/restler-fuzzer/issues/632) that gets generated by the use of double quotes in the log messages in the network.testing.x.logs, therefore the regex has been modified to include double quotes

## PR Checklist
* [x] Applies to work item: #632 
* [] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/onefuzz) and sign.
* [ ] Tests added/passed.  
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different approach. Issue number where discussion took place: #xxx

## Info on Pull Request

This includes only the Regex change

## Validation Steps Performed

Run the results analyzer with a network testing log that has double quotes and single quotes

### DIFF
--- a/src/ResultsAnalyzer/Common/Log.fs
+++ b/src/ResultsAnalyzer/Common/Log.fs
@@ -37,7 +37,7 @@ module Log =
         match line with
         | Regex "^Generation-\d+: Rendering Sequence-\d+" [] ->
             Some SequenceBeginning
-        | Regex "^([^']*): Sending: '(.*)'" [ date; request ] ->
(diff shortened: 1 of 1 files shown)


marina-p:
Hello @Alebmurillo,

I tried this fix on the repro GET request in #632, and did not see any behavior change.  Could you please share the snippet of the network log for which your change resolved the exception for you?

Thanks,

Marina

PR #876, opened 2024-03-14
Enable traceDB-based replay without the request text

marina-p:
For use cases when a trace database is saved and re-played, this change allows using only replay blocks without the rendered requests. This is being added for scenarios where only replay blocks are saved during custom serialization.

The rendered requests continue to be logged by default, since they are useful for debugging.

Testing:
- manual testing
- adding logic to the database replay unit test to create a database without the request text and confirm replay matches the version with the request text.


restler/utils/logging/ndjson_serializer.py
```diff
@@ -120,10 +119,12 @@ def to_dict(self):
...
+        request_text = None if omit_request_text == True else self.request
+        
         return {
             'sent_timestamp': self.sent_timestamp,
             'received_timestamp': self.received_timestamp,
-            'request': self.request,
+            'request': request_text,
             'response': self.response,
             'request_json': None if self.request_json is None else json.dumps(self.request_json),
             'response_json': None if self.request_json is None else json.dumps(self.response_json),
```
wilbaker, line 130: Unrelated to the feature in this PR, I think there might be a bug here:

`'response_json': None if self.request_json is None else json.dumps(self.response_json),`

Should it be:

`: None if self.response_json`

PR #888, opened 2024-04-12
Fix Content-Type custom payload bugs

marina-p:
When a custom payload body is used to work around unsupported content types, the Content-Type header must also be modified via the dictionary.  This functionality had several bugs, which are fixed in this change.  After this update, using either `restler_custom_payload_header` or `restler_custom_payload` with the request-specific syntax for dictionary payloads now works as expected, for example:

  "restler_custom_payload": {
    "/stores/{storeId}/order/post/Content-Type": [ "xml" ]
  }

This change also fixes the same bugs with per-request syntax with restler_custom_payload_query.

Testing:
- manual testing with demo_server
- added new compiler tests

### DIFF
--- a/src/compiler/Restler.Compiler/Compiler.fs
+++ b/src/compiler/Restler.Compiler/Compiler.fs
@@ -698,7 +698,7 @@ module private XMsPaths =
 /// Given a list of the parameters found in the spec and the dictionary,
 /// determines which additional injected parameters are specified in the dictionary
 /// and creates the corresponding payloads.
-let private getInjectedCustomPayloadParameters (dictionary:MutationsDictionary) customPayloadType parametersFoundInSpec =
+let private getInjectedCustomPayloadParameters (dictionary:MutationsDictionary) customPayloadType excludedParameters =
     let parameterNames =
         let parameterNamesSpecifiedAsCustomPayloads =
             match customPayloadType with
@@ -709,7 +709,7 @@ let private getInjectedCustomPayloadParameters (dictionary:MutationsDictionary) 
             | _ ->
                 raise (invalidArg "customPayloadType" (sprintf "%A is not supported in this context." customPayloadType))
         parameterNamesSpecifiedAsCustomPayloads
-        |> Seq.filter (fun name -> not (parametersFoundInSpec |> List.contains name))
+        |> Seq.filter (fun name -> not (excludedParameters |> List.contains name))
 
     parameterNames
         |> Seq.map (fun headerName ->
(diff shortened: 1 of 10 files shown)


src/compiler/Restler.Compiler/Dictionary.fs
```diff
@@ -58,8 +58,18 @@ type MutationsDictionary =
...
             | None -> None
 
+        member x.getRequestTypePayloadPrefix endpoint (method:string) = 
+            sprintf "%s/%s/" endpoint (method.ToLower())
+
+        member x.getRequestTypePayloadName endpoint (method:string) propertyNameOrPath = 
+            sprintf "%s/%s/%s" endpoint (method.ToLower()) propertyNameOrPath
+
+        member x.isRequestTypePayloadName endpoint (method:string) (propertyValue:string) = 
+             let prefix = sprintf "%s/%s" endpoint (method.ToLower())
```
wilbaker, line 68: Should `prefix` include a trailing `/` like it does here in `getRequestTypePayloadPrefix`?

```
        member x.getRequestTypePayloadPrefix endpoint (method:string) = 
            sprintf "%s/%s/" endpoint (method.ToLower())
```

Could there be an issue here with `propertyValue.StartsWith(prefix)` incorrectly matching with a `prefix` that's actually a substring of `propertyValue`?

prefix = "foo/bar"
propertyValue = "foo/bar2"

src/compiler/Restler.Compiler/Compiler.fs
```diff
@@ -858,15 +858,59 @@ let generateRequestPrimitives (requestId:RequestId)
...
-            // Filter out Content-Type, because this is handled separately below
-            let injectedCustomPayloadHeaderParameters = getInjectedCustomPayloadParameters dictionary CustomPayloadType.Header ["Content-Type"]
+            // No headers in Swagger
+            let injectedCustomPayloadHeaderParameters = getInjectedHeaderOrQueryParameters [] CustomPayloadType.Header
             [(ParameterPayloadSource.DictionaryCustomPayload, ParameterList (injectedCustomPayloadHeaderParameters |> seq))]
+        | [(Schema, ParameterList y)] when y |> Seq.isEmpty ->
+            // No headers in Swagger
+            let injectedCustomPayloadHeaderParameters = getInjectedHeaderOrQueryParameters [] CustomPayloadType.Header
+            if injectedCustomPayloadHeaderParameters |> Seq.isEmpty then
+                // TODO: remove this branch and update grammar.py test baselines
```
wilbaker, line 910: Just wanted to check if this TODO is for this PR, or if it's to be addressed in future set of changes.

wilbaker (changes requested):
Wanted to make sure the prefix check in Dictionary.fs is correct

src/compiler/Restler.Compiler/Compiler.fs
marina-p: It's for a future set of changes, updating the grammar.json files is a bit time consuming - in addition to the compiler updates, need to make sure the engine can consume the new grammar.json files and this particular change could introduce a regression.

wilbaker (approved):
Changes look good to me. FYI that I have not tested/validated these changes myself. Thanks!

PR #912, opened 2024-09-11
Omitting authentication tokens from the request text while logging to the trace database files

SwamyNallamalli:
When no_tokens_in_logs is set to true, the RESTler traditional logger detects the authorization tokens in the requests and replaces the same with _OMITTED_AUTH_TOKEN_. The same is not happening when logging is enabled to trace logger files using the setting use_trace_database.
The changes in this pull request aim to copy the functionality to trace logger and make trace logger also omit the auth headers during logging.

### DIFF
--- a/restler/utils/logging/ndjson_serializer.py
+++ b/restler/utils/logging/ndjson_serializer.py
@@ -13,7 +13,7 @@ import logging
 
 from restler_settings import Settings
 from utils.logging.serializer_base import *
-
+import utils.logger as logger
 
 class CustomRotatingFileHandler(RotatingFileHandler):
     def rotation_filename(self, default_name):
@@ -120,6 +120,7 @@ class RequestTraceLog():
         tags.update(self.tags)
         tags.update(self.sequence_tags)
         request_text = None if omit_request_text == True else self.request
+        request_text = logger.remove_tokens_from_logs(request_text) if Settings().no_tokens_in_logs else request_text
         
         return {
             'sent_timestamp': self.sent_timestamp,
--- a/restler/utils/logging/ndjson_serializer.py
+++ b/restler/utils/logging/ndjson_serializer.py
@@ -106,7 +106,7 @@ class RequestTraceLog():
     def response(self, value):
         self._response = value
 
-    def to_dict(self, omit_request_text=None):
+    def to_dict(self, omit_request_text=None, remove_tokens_from_logs=True):
         tags = {}
         if self.request_id is not None:
             tags["request_id"] = self.request_id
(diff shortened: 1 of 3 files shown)


restler/utils/logging/ndjson_serializer.py
```diff
@@ -120,6 +120,7 @@ def to_dict(self, omit_request_text=None):
         tags.update(self.tags)
         tags.update(self.sequence_tags)
         request_text = None if omit_request_text == True else self.request
+        request_text = logger.remove_tokens_from_logs(request_text) if Settings().no_tokens_in_logs else request_text
```
marina-p, line 123: >Settings().no_tokens_in_logs  [](http://example.com/codeflow?start=71&length=29)

Hi Swamy, thanks for the fix!

The check-in tests did not run because we need to migrate them to GitHub actions (this will be done soon), so you will need to run unit tests locally.  For this change, you can just run 'test_trace_database*' tests for validation.

The Settings() object can't be used directly here - could you please restructure the code as follows, which should fix the issue:

```
    def to_dict(self, omit_request_text=None, remove_tokens_from_logs=True):
        from utils.logger import remove_tokens_from_logs as remove_tokens
        tags = {}
...
        request_text = None if omit_request_text == True else self.request
        request_text = remove_tokens(request_text) if remove_tokens_from_logs else request_text 
```

then just one place in trace_db.py line 202 needs to be updated with the new parameter set to the setting.

marina-p (changes requested):
:clock1:

restler/utils/logging/ndjson_serializer.py
SwamyNallamalli: Thanks for detailed code suggestions.
Somehow I saw the usage of Settings() in the file and assumed it will work.
I did run unit tests locally and they failed, but I ran all of those and it took some time. 
Now the tests are all succeeding. 

Is there a place I should add verification in the unit tests to check if auth tokens are actually getting replaced or not?

SwamyNallamalli:
@microsoft-github-policy-service agree company="Microsoft"

restler/utils/logging/ndjson_serializer.py
marina-p: Hi Swamy, thanks for updating the test.  For validating the token replacement was actually performed, sanity checking for the presence of the OMITTED_TOKEN constant (reading the trace DB file directly without the deserializer) is sufficient. For example, counting the number of requests and making sure there are that many occurrences of this constant).

marina-p (changes requested):
:clock1:

marina-p:
The update looks good!  Left just one comment RE: the test update

restler/utils/logging/ndjson_serializer.py
SwamyNallamalli: Had to update the unit test to see the impact of the flag and validated that authoization token is getting replaced.

marina-p (approved):
:shipit:

PR #915, opened 2024-11-13
Add validation to engine dictionary

### LINKED ISSUE #773: The engine should validate the dictionary values are all quoted
### 💡 Idea

The array values in the dictionary must all be strings.  For example:

Correct way to define a fuzzable int:
```
  "restler_fuzzable_int": [
    "1"
  ],

```

Incorrect way to define a fuzzable int:
```
  "restler_fuzzable_int": [
    1
  ],

```
We need to validate this in the engine before proceeding, since later errors can be difficult to debug.

### Design Notes

_No response_

brokoli777:
Fixes #773

This adds a local function to see if `custom_values` dictionary only contains strings. It will throw an error if the dictionary is not valid.
I have also added a test case for testing the code change.

Please let me know if there are any changes to be made, as it is my first time contributing to this codebase.

### DIFF
--- a/restler/unit_tests/test_restler_settings.py
+++ b/restler/unit_tests/test_restler_settings.py
@@ -235,6 +235,30 @@ class RestlerSettingsTest(unittest.TestCase):
         self.assertEqual(None, settings.get_checker_arg('leakage_rule_mode', 'mode'))
         self.assertEqual(None, settings.get_checker_arg('resource_hierarchy_rule_mode', 'mode'))
 
+    def test_validate_string_values_in_candidate_values(self):
+        candidate_values = primitives.CandidateValuesPool()
+        
+        invalid_dict = {
+            "restler_fuzzable_int": [
+                1  # Invalid as it's not a string
+            ]
+        }
+
+        with self.assertRaises(ValueError):
+            candidate_values.set_candidate_values(invalid_dict)
+
+        valid_dict = {
+            "restler_fuzzable_int": [
+                "1"  # Valid as it's a string
+            ]
+        }
+
+        # No error should since input is valid
+        try:
+            candidate_values.set_candidate_values(valid_dict)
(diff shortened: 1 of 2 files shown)


brokoli777:
Unfortunately, I couldn't run the unit tests due to the following error:

image

I have installed the packages from requirements.txt

marina-p:
Hi, thank you for adding this validation.  The PR and test looks good.  We've had CI checks broken for a few weeks, I just re-enabled them and everything including the new test is passing, so I'll go ahead and merge this: [https://github.com/microsoft/restler-fuzzer/actions/runs/12292021287](https://github.com/microsoft/restler-fuzzer/actions/runs/12292021287).  For troubleshooting running tests locally, what was the exact command/directory/environment?  

---
In reply to: [2473982009](https://github.com/microsoft/restler-fuzzer/pull/915#issuecomment-2473982009) [](http://example.com/codeflow?ancestors=2473982009)

marina-p (dismissed):
:shipit:

restler/unit_tests/test_restler_settings.py
```diff
@@ -235,6 +235,30 @@ def test_checkers_in_settings(self):
         self.assertEqual(None, settings.get_checker_arg('leakage_rule_mode', 'mode'))
         self.assertEqual(None, settings.get_checker_arg('resource_hierarchy_rule_mode', 'mode'))
 
+    def test_validate_string_values_in_candidate_values(self):
+        candidate_values = primitives.CandidateValuesPool()
```
marina-p, line 239: >candidate_values [](http://example.com/codeflow?start=8&length=16)

This is failing since the Settings object is not initialized.  Each test needs to have the following initialization: ```        settings_file = {}
_ = RestlerSettings(settings_file)
```

marina-p (approved):
:shipit:

PR #960, opened 2025-03-14
Update README with away notice

marina-p:


machyonParticle (approved):
Best wishes on your leave!

PR #972, opened 2025-06-23
Add Dependabot configuration for .NET SDK updates

JamieMagee:
This PR adds Dependabot configuration to automatically manage .NET SDK version updates in global.json.

This configuration will:
- Monitor the global.json file for .NET SDK version updates
- Create pull requests when new SDK versions are available
- Help keep the project secure with the latest SDK patches

For more information, see: https://devblogs.microsoft.com/dotnet/using-dependabot-to-manage-dotnet-sdk-updates/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "dotnet-sdk"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    ignore:
+      - dependency-name: "*"
+        update-types: 
+          - "version-update:semver-major"
+          - "version-update:semver-minor"


marina-p:
I'd like to configure this to just focus on security updates - please let me know if you see any issues.

```
version: 2
updates:
  - package-ecosystem: "dotnet-sdk"
    directory: "/"
    schedule:
      interval: "weekly"
      day: "wednesday"
    # Only allow security updates, ignore everything else
    open-pull-requests-limit: 10
    ignore:
      - dependency-name: "*"
        update-types: 
          - "version-update:semver-major"
          - "version-update:semver-minor"
          - "version-update:semver-patch"
    # Enable security updates explicitly
    security-updates-only: true
```

JamieMagee:
Unfortunately, the Dependabot .NET SDK updater only supports version updates, not security updates. However, it's mostly a moot point as ~90% of updates are security related anyway. Check out the CVEs fixed in each .NET 8 SDK release for example: https://github.com/dotnet/core/blob/main/release-notes/8.0/cve.md

